PR #13, opened 2020-06-19
Prevent early deallocation of object references

MaxDesiatov:
Currently, when two references are created for the same object (i.e. with the same `id`), deallocating the first references invalidates the second. This is can cause crashes, which is reproduced in the newly added test. To avoid this issue, a new private `cache` of weak references is created, which can retrieve the same `JSObjectRef` instance for the same id, when called from `RawJSValue.jsValue()`.

Obviously, this adds some overhead on object retrieval when the Swift <-> JS bridge is crossed, but I think invalid references are a much more serious problem than this overhead. I don't know yet what implementation could be more efficient, if that's possible at all.

### DIFF
--- a/IntegrationTests/JavaScriptKitExec/Sources/JavaScriptKitExec/main.swift
+++ b/IntegrationTests/JavaScriptKitExec/Sources/JavaScriptKitExec/main.swift
@@ -299,3 +299,27 @@ Object_Conversion: do {
 } catch {
     print(error)
 }
+
+ObjectRef_Lifetime: do {
+    // ```js
+    // global.globalObject1 = {
+    //   "prop_1": {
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
I think it should be a better approach to manage a reference count in JavaScript library https://github.com/kateinoigakukun/JavaScriptKit/pull/14
What do you think about the approach?

MaxDesiatov:
Yeah, that could work too. I wasn't sure if doing anything on the JavaScript side is slower, but most probably just incrementing and decrementing a counter is faster than maintaining a dictionary with its hashing algorithms 🙂 Also, could be more memory efficient, since you don't waste the space on the cache dictionary

Issue #21, opened 2020-07-22
DOMKit?


j-f1:
It would be great to have some sort of type-safe DOM API that better matches DOM as exposed to JS (for example `document.querySelector('foo')?.appendChild(document.createElement('bar'))`) without having to go through the ceremony of `.object!` and `.function!`. It looks like there’s some work in this direction in [`webidl2swift`](https://github.com/Apodini/webidl2swift) but I’m not sure how much more is needed.

MaxDesiatov:
Generating the `WebAPI` module as described [in the `webidl2swift` example](https://github.com/Apodini/webidl2swift/tree/develop/Example) worked for me, but they have a dependency on [a fork of JavaScriptKit](https://github.com/Unkaputtbar/JavaScriptKit/commit/8f0a15464238781a90141b93558d79c14d101bd5). I think upstreaming changes from that fork could be the first step.

j-f1:
I tried to merge but it seems like they made a bunch of unrelated changes in the fork which seem vaguely like a good idea since they reduce memory usage (I think) and cut down on the number of arguments that need to be passed around on the JS side but that means it’s hard to extract just the new stuff

j-f1:
I’ve got DOMKit building! (haven’t tried running it yet)

https://github.com/j-f1/DOMKit

```swift
let document = DOMKit.Document(jsValue: JSObjectRef.global.document)

let button = document.createElement(localName: "button")
button.textContent = "Hello, world"

class ClickListener: EventListener {
    required init(objectRef: JSObjectRef) {
        self.objectRef = objectRef
    }

    var objectRef: JSObjectRef
    static func canDecode(from jsValue: JSValue) -> Bool { false }
    func handleEvent(event: Event) {
        (event.target as? HTMLElement)?.textContent = "Clicked!"
    }
}

let emptyObject = JSObjectRef.global.Object.function!(.new)

button.addEventListener(type: "click", callback: ClickListener(objectRef: emptyObject))
_ = document.querySelector(selectors: "body")?.appendChild(node: button)
```

j-f1:
[...]

PR #22, opened 2020-08-02
Webidl support

j-f1:
Progress on #21. Currently broken and I might start over

It seems like the 2 most important changes are:

- `instanceof` support (which is implemented hackily but is ultimiately necessary for properly type checking things)
  - I kinda want to re-implement this in a way that actually lets you provide the class object instead of pulling it off `window`.
- JSValue encoding/decoding to seamlessly pass in and return Swift values

Unrelated changes:
- instead of having type + 3 payload slots per object, have just type + two slots and interpret them as a combined `double` or two 32-bit ints depending on the type
  - this would cut down memory usage a bunch although I don‘t think it would really be a significant problem for a while
- instead of passing around `kind, payload1, payload2, payload3`, pass around a `DataView` into the memory object, offset to the right location
  - this makes it easier to pass around blobs of memory I guess, but not sure what the perf implications are


kateinoigakukun:
@j-f1 Awesome progress! If you break this big change down into several small PRs, it would make it easy to review for me :pray:

j-f1:
Will do! Closing in favor of #24 and future PRs

PR #24, opened 2020-08-02
Pass around a DataView instead of individual kind/payload fields

j-f1:
Also change memory layout for numbers:

There are now only two payload slots. Numbers use them as a single Double, while every other type just uses them as two UInt32s

Originally part of #22.

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -247,8 +247,7 @@ export class SwiftRuntime {
         }
 
         // Note:
-        // `decodeValues` assumes that the size of RawJSValue is 24
-        // and the alignment of it is 8
+        // `decodeValues` assumes that the size of RawJSValue is 12
         const decodeValues = (ptr: pointer, length: number) => {
             let result = []
             for (let index = 0; index < length; index++) {
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -180,7 +180,8 @@ export class SwiftRuntime {
                       dataView.getUint32(payload2_ptr, true)
                     );
                 }
-                case JavaScriptValueKind.Object: {
+                case JavaScriptValueKind.Object:
+                case JavaScriptValueKind.Function: {
                     return this.heap.referenceHeap(dataView.getUint32(payload1_ptr, true))
                 }
                 case JavaScriptValueKind.Null: {
@@ -188,9 +189,6 @@ export class SwiftRuntime {
                 }
                 case JavaScriptValueKind.Undefined: {
                     return undefined
-                }
-                case JavaScriptValueKind.Function: {
-                    return this.heap.referenceHeap(dataView.getUint32(payload1_ptr, true))
                 }
                 default:
                     throw new Error(`Type kind "${kind}" is not supported`)
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -155,24 +155,9 @@ export class SwiftRuntime {
             }
         }
 
-        const readUInt32 = (ptr: pointer) => {
-            const uint32Memory = new Uint32Array(memory().buffer, ptr, 1);
-            return uint32Memory[0];
-        }
-
-        const readFloat64 = (ptr: pointer) => {
-            const dataView = new DataView(memory().buffer);
-            return dataView.getFloat64(ptr, true);
-        }
-
         const writeUint32 = (ptr: pointer, value: number) => {
             const uint32Memory = new Uint32Array(memory().buffer, ptr);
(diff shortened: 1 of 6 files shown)


j-f1:
Note: I dropped the size of each object in memory from 24 down to 12. I’m not exactly sure why they were 24 bytes long before since kind + payload1 + payload2 = 12 bytes and payload3 = 8 bytes which only add up to 20. If I’m missing something about the memory layout, I’d be happy to correct it.

Runtime/src/index.ts
```diff
@@ -124,10 +124,12 @@ export class SwiftRuntime {
             const exports = this.instance.exports as any as SwiftRuntimeExportedFunctions;
             const argc = args.length
             const argv = exports.swjs_prepare_host_function_call(argc)
+            const uint32Memory = new Uint32Array(memory().buffer, argv, args.length * 3)
```
kateinoigakukun, line 127: This `uint32Memory` is not used

kateinoigakukun:
Thanks for your cooperation.

Did you check performance benchmark suite? You can check benchmark by `make benchmark` in `IntegrationTests` directory.

I found some performance regression on number interoperation. 

Before

```
Running 'Serialization/Write JavaScript number directly' ...
done 118.31193709373474 ms
Running 'Serialization/Write JavaScript string directly' ...
done 119.49532198905945 ms
Running 'Serialization/Swift Int to JavaScript' ...
done 3457.1316990852356 ms
Running 'Serialization/Swift String to JavaScript' ...
done 5258.867609977722 ms
Running 'Object heap/Increment and decrement RC' ...
done 3061.0713909864426 ms
```

After
```
Running 'Serialization/Write JavaScript number directly' ...
done 118.58276605606079 ms
Running 'Serialization/Write JavaScript string directly' ...
done 120.5656269788742 ms
Running 'Serialization/Swift Int to JavaScript' ...
done 4948.468513011932 ms
Running 'Serialization/Swift String to JavaScript' ...
done 5090.137539982796 ms
Running 'Object heap/Increment and decrement RC' ...
done 2963.7367210388184 ms
```


I'm still reviewing the changes, so please wait for a while.

kateinoigakukun:
> I’m not exactly sure why they were 24 bytes long before since kind + payload1 + payload2 = 12 bytes and payload3 = 8 bytes which only add up to 20. 

The additional 4 byte is for `double` which is aligned by 8 bytes in C convention

kateinoigakukun:
I think the regression is due to the indirection of payloads access.
Why don't you pass the two payloads directly as arguments?

j-f1:
> I think the regression is due to the indirection of payloads access.

That definitely seems to be the case! I’m not sure why they switched to passing in the `DataView`s but I’d be happy to revert that part

> The additional 4 byte is for `double` which is aligned by 8 bytes in C convention

Ah, that makes sense! Since we’re not using C doubles anymore (we’re just encoding them as two ints), this shouldn’t be an issue, right?

kateinoigakukun:
> That definitely seems to be the case! I’m not sure why they switched to passing in the DataViews but I’d be happy to revert that part

Thanks ❤️ 

> Ah, that makes sense! Since we’re not using C doubles anymore (we’re just encoding them as two ints), this shouldn’t be an issue, right?

Yes, I think it's ok.

j-f1:
Ok, so I did a test where I removed all the `DataView` stuff (except when decoding numbers (where we need to specify endianness, I think?)) but keeping the new function style where we only pass a single pointer representing the entire JSValue rather than `kind`/`payload1`/`payload2` and it’s somehow slower than all the other styles?

| Test | `master` | single pointer | both |
|-|-|-|-|
| Serialization/Write JavaScript number directly | 123 | 126 | 122 |
| Serialization/Write JavaScript string directly | 130 | 134 | 126 |
| Serialization/Swift Int to JavaScript | 3150 | 5075 | 5083 |
| Serialization/Swift String to JavaScript | 4842 | 4925 | 4838 |
| Object heap/Increment and decrement RC | 2898 | 2817 | 2904 |

j-f1:
Ultimately I’d be fine just closing this if it has no real benefit

kateinoigakukun:
Hmm, the benchmark results show that decoding number from two 32bit payloads is slower than passing 20 byte payloads directly.

I think we need to optimize the payloads, but this approach seems a little difficult.

j-f1:
This might be a good thing to explore in the future when there’s better tooling to analyze perf but I don’t have a good rationale for this PR right now. Not really sure why I opened it in the first place. Thanks for sticking with me as I figure out exactly which things need to be brought over!

PR #25, opened 2020-08-02
Add support for instanceof

j-f1:
~This depends on #24. I’ll rebase when that’s done.~

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -59,7 +59,7 @@ class SwiftRuntimeHeap {
 
     allocHeap(value: any) {
         const isObject = typeof value == "object";
-        const entry = this._heapEntryByValue.get(value); 
+        const entry = this._heapEntryByValue.get(value);
         if (isObject && entry) {
             entry.rc++
             return entry.id
@@ -368,8 +368,7 @@ export class SwiftRuntime {
             ) => {
               const obj = this.heap.referenceHeap(obj_ref)
               const constructor = this.heap.referenceHeap(constructor_ref)
-              const result = obj instanceof constructor
-              writeUint32(result_ptr, Number(result))
+              return obj instanceof constructor
             },
             swjs_destroy_ref: (ref: ref) => {
                 this.heap.freeHeap(ref)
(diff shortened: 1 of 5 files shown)


j-f1:
Update: I was able to separate this completely from #24

Runtime/src/index.ts
```diff
@@ -362,6 +362,15 @@ export class SwiftRuntime {
...
                 writeUint32(result_obj, this.heap.allocHeap(result));
             },
+            swjs_instanceof: (
+                obj_ref: ref, constructor_ref: ref,
+                result_ptr: pointer
+            ) => {
+              const obj = this.heap.referenceHeap(obj_ref)
+              const constructor = this.heap.referenceHeap(constructor_ref)
+              const result = obj instanceof constructor
+              writeUint32(result_ptr, Number(result))
```
kateinoigakukun, line 372: Why don't you just return the result value?
```suggestion
              return obj instanceof constructor
```

Sources/JavaScriptKit/JSObject.swift
```diff
@@ -32,6 +32,16 @@ public class JSObjectRef: Equatable {
         getJSValue(this: self, index: Int32(index))
     }
 
+    public func instanceof(_ constructor: JSValue) -> Bool {
+        instanceof(constructor.function!)
```
kateinoigakukun, line 36: This library shouldn't provide unsafe method

Runtime/src/index.ts
j-f1: Oh, cool! Wasn’t sure if this was possible

Sources/JavaScriptKit/JSObject.swift
j-f1: I changed it to default to `false` if you don’t pass a function in. Does that make sense?

kateinoigakukun: I think this convenience method is unnecessary. The users should handle the non-function case.

MaxDesiatov: I agree, `new` is similarly defined on `JSFunctionRef`, not `JSObjectRef` IIRC

PR #26, opened 2020-08-02
Add JSBridgedType and JSBridgedClass

j-f1:
Ref #21.

This allows DOMKit to build, and implements `JSBridgedClass` on `JSError` and `JSDate`.


Sources/JavaScriptKit/JSBridgedType.swift
```diff
@@ -0,0 +1,31 @@
...
+    public func jsValue() -> JSValue { value }
+
+    public var description: String { value.description }
+}
+
+
+public protocol JSBridgedClass: JSBridgedType {
+    static var constructor: JSFunction { get }
+    var jsObject: JSObject { get }
+    init(withCompatibleObject jsObject: JSObject)
```
j-f1, line 22: I feel like this should have `unsafe` in the name but I’m unsure of how to word it.

Sources/JavaScriptKit/Support.swift
```diff
@@ -0,0 +1,57 @@
+protocol _AnyJSValueCodable: JSValueConvertible {}
+
+public struct AnyJSValueCodable: JSValueCodable, ExpressibleByNilLiteral {
```
j-f1, line 3: This is necessary for things like [`AddEventListenerOptions`](https://github.com/j-f1/DOMKit/blob/main/Sources/DOMKit/WebIDL/AddEventListenerOptions.swift).

MaxDesiatov:
It's still a bit unclear how exactly DOMKit uses `JSBridgedType` and `JSBridgedClass`. It adds conformance on those protocols to its types, but could you please elaborate how exactly it's used? I initially thought it's needed for `dynamicCast` and `staticCast` functions, but I don't think they're used anywhere in `DOMKit`.

j-f1:
It simplifies the class implementation. For example, `Event` specifies its `constructor` function (`JSObject.global.EventTarget.function!`), and then JavaScriptKit is able to implement `construct(from: JSValue)` automatically, checking that the passed value is an object that is `instanceof` the constructor. It also provides a good starting point for people trying to bridge a class over from JS.

What are `staticCast` and `dynamicCast` used for? I couldn’t find any info about them.

MaxDesiatov:
Makes sense. 👍 

IDK if this PR is ready for review, but could `staticCast` and `dynamicCast` be removed then? Also, with `swift-doc` deployment it would be great for all new public APIs to get doc comments.

Sources/JavaScriptKit/JSValueConvertible.swift
```diff
@@ -79,24 +95,80 @@ extension Dictionary: JSValueConvertible where Value == JSValueConvertible, Key
...
             array[index] = element.jsValue()
         }
         return .object(array)
     }
 }
 
+extension Array: JSValueConstructible where Element: JSValueConstructible {
+    public static func construct(from value: JSValue) -> [Element]? {
+        if let objectRef = value.object,
+           objectRef.isInstanceOf(JSObject.global.Array.function!) {
```
kateinoigakukun, line 156: ditto ⬆️

Sources/JavaScriptKit/JSValueConvertible.swift
```diff
@@ -79,24 +95,80 @@ extension Dictionary: JSValueConvertible where Value == JSValueConvertible, Key
     }
 }
 
-private let Array = JSObject.global.Array.function!
+private let NativeJSArray = JSObject.global.Array.function!
+extension Dictionary: JSValueConstructible where Value: JSValueConstructible, Key == String {
+    public static func construct(from value: JSValue) -> Self? {
+        if let objectRef = value.object,
```
kateinoigakukun, line 101: It would be more clear if you use `guard` instead.

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -89,6 +90,15 @@ public class JSClosure: JSFunction {
...
 
+    public required convenience init(jsValue: JSValue) {
+        switch jsValue {
+        case let .function(fun):
+            self.init { fun(arguments: $0) }
+        default:
+            fatalError()
+        }
+    }
+
```
kateinoigakukun, line 101:
```suggestion
    public override class func construct(from value: JSValue) -> Self? {
        guard case let .function(fun) = value else { return nil }
        return JSClosure { fun(arguments: $0) } as? Self
    }

```

kateinoigakukun (commented):
Looks good but I added some comments

Sources/JavaScriptKit/Support.swift
```diff
@@ -0,0 +1,43 @@
+protocol _AnyJSValueConvertible: JSValueConvertible {}
+
+/// A type-erased wrapper for types conforming to JSValueConvertible
+public struct AnyJSValueConvertible: JSValueCodable, ExpressibleByNilLiteral {
```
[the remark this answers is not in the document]
j-f1: Screenshot_2020-09-15 17 55 35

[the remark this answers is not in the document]
kateinoigakukun: In that case, users should use JSValue instead of `JSValueCodable`.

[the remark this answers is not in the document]
j-f1: I’ve found one more place `AnyJSValueConvertible` comes up: APIs like [`console`](https://github.com/swiftwasm/DOMKit/blob/main/Sources/DOMKit/WebIDL/console.swift) that accept any value. I think it is possible to get this to work but it would require a redesign of a bunch of generated code.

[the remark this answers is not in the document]
kateinoigakukun: Hmm, I think the arguments of `console` methods can be replaced with `JSValue` also

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
j-f1: This is already implemented here:

https://github.com/j-f1/forked-JavaScriptKit/blob/479ad20b07f36a2db1cb5a410f12a7ae587d1b49/Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift#L57-L59

Should I remove this `init(jsValue:)` declaration then?

kateinoigakukun: Yes, it would be more consistent interface.

Sources/JavaScriptKit/Support.swift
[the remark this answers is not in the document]
j-f1: Yep! There was just one case with an `any` in the WebIDL files that had a default `nil` value which should be cleared up by #59.

Sources/JavaScriptKit/JSValue.swift
```diff
@@ -57,12 +57,21 @@ public enum JSValue: Equatable {
...
+        return self == .null
+    }
 
     /// Returns the `true` if this JS value is undefined.
     /// If not, returns `false`.
-    public var isUndefined: Bool { return self == .undefined }
+    public var isUndefined: Bool {
+        return self == .undefined
+    }
+}
```
j-f1, line 69: Now that I’m reviewing this section — how are these useful? Isn’t it better to use one of the other type accessors?

It looks like they’re used in `JSValueDecoder` and nowhere else.

MaxDesiatov: I concur, but Yuta should make the call 🙂

kateinoigakukun: I'm thinking changing JSValue to be struct instead enum to be more resilient interface. And for future coming JSString implementation. 

If changing from enum to struct, value == .null will be broken. So I may restore this in the future. But for now, it's ok to remove them.

j-f1: If you add `static let null = ...` to your new JSValue implementation, `value == .null` will still work!

Sources/JavaScriptKit/JSBridgedType.swift
MaxDesiatov: `withUnsafeCast`?

kateinoigakukun:
Please let me know when it's ready for review again 👍

Sources/JavaScriptKit/BasicObjects/JSArray.swift
```diff
@@ -1,54 +1,63 @@
...
     public typealias Element = JSValue
 
     public func makeIterator() -> Iterator {
-        Iterator(ref: ref)
+        Iterator(jsObject: jsObject)
     }
 
     public class Iterator: IteratorProtocol {
-        let ref: JSObject
+        let jsObject: JSObject
```
MaxDesiatov, line 39:
```suggestion
        private let jsObject: JSObject
```

Sources/JavaScriptKit/BasicObjects/JSArray.swift
```diff
@@ -1,54 +1,63 @@
...
         var index = 0
```
MaxDesiatov, line 40:
```suggestion
        private var index = 0
```

Sources/JavaScriptKit/JSBridgedType.swift
```diff
@@ -0,0 +1,41 @@
+/// Use this protocol when your type has no single JavaScript class.
+/// For example, a union type of multiple classes or primitive values.
```
MaxDesiatov, line 2: Could this comment give an example of such type?

Sources/JavaScriptKit/JSBridgedType.swift
```diff
@@ -0,0 +1,41 @@
...
+    public var description: String { value.description }
+}
+
+/// Conform to this protocol when your Swift class wraps a JavaScript class.
+public protocol JSBridgedClass: JSBridgedType {
+    /// The constructor function for the JavaScript class
+    static var constructor: JSFunction { get }
+
+    /// The JavaScript object wrapped by this instance.
+    /// You may assume that `jsObject instanceof Self.constructor`
```
MaxDesiatov, line 27:
```suggestion
    /// You may assume that `jsObject instanceof Self.constructor == true`
```

Sources/JavaScriptKit/JSBridgedType.swift
MaxDesiatov: Fine by me 👍

Sources/JavaScriptKit/JSValue.swift
```diff
@@ -138,3 +153,37 @@ public func setJSValue(this: JSObject, index: Int32, value: JSValue) {
                        rawValue.payload1, rawValue.payload2, rawValue.payload3)
     }
 }
+
+extension JSValue {
+    public func isInstanceOf(_ constructor: JSFunction) -> Bool {
```
MaxDesiatov, line 158: Is a doc comment for this available in some protocol that has `isInstanceOf` requirement (I don't remember if there is one off the top of my head)? If not, then it needs a doc comment as a `public` declaration.

Sources/JavaScriptKit/JSValueConvertible.swift
```diff
@@ -81,24 +95,82 @@ extension Dictionary: JSValueConvertible where Value == JSValueConvertible, Key
     }
 }
 
-private let Array = JSObject.global.Array.function!
+private let NativeJSArray = JSObject.global.Array.function!
```
MaxDesiatov, line 98: **nit:** the convention is that only Swift types start with an uppercase.

```suggestion
private let arrayConstructor = JSObject.global.Array.function!
```

Sources/JavaScriptKit/JSValueConvertible.swift
```diff
@@ -194,6 +266,6 @@ extension Array where Element == JSValueConvertible {
 
 extension Array where Element: JSValueConvertible {
     func withRawJSValues<T>(_ body: ([RawJSValue]) -> T) -> T {
-        Swift.Array<JSValueConvertible>.withRawJSValues(self)(body)
+        Array<JSValueConvertible>.withRawJSValues(self)(body)
```
MaxDesiatov, line 269: This would also work if you like conciseness 🙂

```suggestion
        [JSValueConvertible].withRawJSValues(self)(body)
```

Sources/JavaScriptKit/JSBridgedType.swift
j-f1: Sure. Would it be OK to link to DOMKit?￼

Sources/JavaScriptKit/JSValue.swift
j-f1: Nope, just forgot to document it.

Sources/JavaScriptKit/JSBridgedType.swift
j-f1: (Here’s an example: https://github.com/swiftwasm/DOMKit/blob/main/Sources/DOMKit/WebIDL/FileOrString.swift)

MaxDesiatov: Yes, please do 👍

Sources/JavaScriptKit/FundamentalObjects/JSObject.swift
```diff
@@ -62,9 +69,9 @@ public class JSObject: Equatable {
         set { setJSValue(this: self, index: Int32(index), value: newValue) }
     }
 
-    /// Return `true` if this object is an instance of the `constructor`. Return `false`, if not.
+  /// Return `true` if this value is an instance of the passed `constructor` function.
```
MaxDesiatov, line 72: **nit:** some kind of indentation misalignment here?

```suggestion
    /// Return `true` if this value is an instance of the passed `constructor` function.
```

MaxDesiatov (approved):
Fantastic 👏

Just a minor indentation issue, but otherwise LGTM 👍

kateinoigakukun (approved):

Sources/JavaScriptKit/FundamentalObjects/JSObject.swift
j-f1: Indeed. I wish Xcode supported editorconfig so I wouldn’t have indentation mismatches

PR #28, opened 2020-08-03
Clean up the JSObjectRef API

### LINKED ISSUE #27: Separate namespaces for methods and properties?
Over time, it’s likely that more and more methods will be added to JSObjectRefs. Currently, there are `get`, `set`, `jsValue`, `subscript`, and (shortly) `instanceof`. However, at least some of these methods are likely to appear in regular JS objects (specifically `get` and `set`). While there is the escape hatch of calling `object[dynamicMember: "get"]` or `object.get("get")`, this is less than optimal and it would be great if there was a way to have a separate namespace for the JavaScriptKit-defined  methods so they’d never clash with something from JS.

Maybe it would be sufficient to just add a regular `subscript` method `object["get"]` or `object[1]`?

j-f1:
Fixes #27.

- `get`/`set` → `dynamicMember` and subscripts
- `instanceof` (kept)
- `jsValue` (kept)
- `apply` → `callAsFunction`

### DIFF
--- a/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
+++ b/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
@@ -281,12 +281,12 @@ Call_Function_With_This: do {
 }
 
 Object_Conversion: do {
-//    let array1 = [1, 2, 3]
-//    let jsArray1 = array1.jsValue().object!
-//    try expectEqual(jsArray1.length, .number(3))
-//    try expectEqual(jsArray1[0], .number(1))
-//    try expectEqual(jsArray1[1], .number(2))
-//    try expectEqual(jsArray1[2], .number(3))
+   let array1 = [1, 2, 3]
+   let jsArray1 = array1.jsValue().object!
+   try expectEqual(jsArray1.length, .number(3))
+   try expectEqual(jsArray1[0], .number(1))
+   try expectEqual(jsArray1[1], .number(2))
+   try expectEqual(jsArray1[2], .number(3))
 
     let array2: [JSValueConvertible] = [1, "str", false]
     let jsArray2 = array2.jsValue().object!
@@ -296,9 +296,9 @@ Object_Conversion: do {
     try expectEqual(jsArray2[2], .boolean(false))
     _ = jsArray2.push!(5)
     try expectEqual(jsArray2.length, .number(4))
-//    _ = jsArray2.push!(jsArray1)
+   _ = jsArray2.push!(jsArray1)
 
-//    try expectEqual(jsArray2[4], .object(jsArray1))
+   try expectEqual(jsArray2[4], .object(jsArray1))
 
     let dict1: [String: JSValueConvertible] = [
         "prop1": 1,
--- a/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
+++ b/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
@@ -271,14 +271,14 @@ Call_Function_With_This: do {
 
 Object_Conversion: do {
 //    let array1 = [1, 2, 3]
-//    let jsArray1 = JSValue(from: array1).object!
+//    let jsArray1 = array1.jsValue().object!
 //    try expectEqual(jsArray1.length, .number(3))
 //    try expectEqual(jsArray1[0], .number(1))
 //    try expectEqual(jsArray1[1], .number(2))
 //    try expectEqual(jsArray1[2], .number(3))
 
     let array2: [JSValueConvertible] = [1, "str", false]
-    let jsArray2 = JSValue(from: array2).object!
+    let jsArray2 = array2.jsValue().object!
     try expectEqual(jsArray2.length, .number(3))
     try expectEqual(jsArray2[0], .number(1))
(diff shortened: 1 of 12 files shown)


j-f1:
Inspired by the `ExpressibleByNilLiteral` protocol, I came up with this subscript-based `JSValueConvertible` protocol:

```swift
public protocol JSValueConvertible {
    subscript(jsValue _: ()) -> JSValue { get }
}

extension JSValue {
    public init(from convertible: JSValueConvertible) {
        self = convertible[jsValue: ()]
    }
}
```

j-f1:
Alright, the tests are still failing because it seems like I can’t have `Array` conform to `JSValueConvertible` when either `Element: JSValueConvertible` or `Element == JSValueConvertible` :(

Any suggestions?

j-f1:
Update: I’ve replaced all the function call/new logic with this pattern:

```swift
// JS version in comment
func() // func.apply(null, [])
func(arg1, arg2, arg3) // func.apply(null, [arg1, arg2, arg3])
func(args: [arg1, arg2, arg3]) // func.apply(null, [arg1, arg2, arg3])
func(this: foo) // func.apply(foo, [])
func(this: foo, arg1, arg2, arg3) // func.apply(foo, [arg1, arg2, arg3])
func(this: foo, args: [arg1, arg2, arg3]) // func.apply(foo,  [arg1, arg2, arg3])
func(new: arg1, arg2, arg3) // new func(arg1, arg2, arg3)
func(.new, args: [arg1, arg2, arg3]) // new func(arg1, arg2, arg3)
func(.new) // new func()
```

The last two I’m a little uncertain of.

j-f1:
One last change I was going to make was changing the `instanceof` function to an operator. I tried it out with `<<` which isn’t super clear.

It seems Swift doesn’t allow letters in operators, so `instanceof` is out. There are a ton of Unicode characters allowed but the only ASCII ones are `!%&*+-/<=>?^|~`. Maybe `<<?` or `<==`?

kateinoigakukun:
Sorry, I didn't take a look much now, but could you ignore .swiftpm directory?

kateinoigakukun:
I agree with deprecating get and set and provide subscript. 

But on second thought, `new` and `instanceof` are not natural and not swifty interface.

This library provides JavaScript like APIs but also they should be swifty.

I think `instanceof`, `new` and `jsValue` are not common naming as members and if users want to use those names, they can use subscript accessor instead.

I disagree that APIs become unnatural caring about edge cases.

j-f1:
Good point! I reverted those changes, so now the last few samples look like this:

```js
foo.new(arg1, arg2, arg3)
new foo(arg1, arg2, arg3)

foo.new(args: [arg1, arg2, arg3])
new foo(...[arg1, arg2, arg3])

foo.new()
new foo()
```

Additionally, I changed the JSValue code back to using `.jsValue()` and changed `instanceof` to add the method `isInstanceOf`.

IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
```diff
@@ -296,9 +296,9 @@ Object_Conversion: do {
     try expectEqual(jsArray2[2], .boolean(false))
     _ = jsArray2.push!(5)
     try expectEqual(jsArray2.length, .number(4))
-    _ = jsArray2.push!(jsArray1)
+//    _ = jsArray2.push!(jsArray1)
 
-    try expectEqual(jsArray2[4], .object(jsArray1))
+//    try expectEqual(jsArray2[4], .object(jsArray1))
```
kateinoigakukun, line 301: Could you revert these comments out?

Sources/JavaScriptKit/JSFunction.swift
```diff
@@ -1,45 +1,44 @@
 import _CJavaScriptKit
 
-@dynamicCallable
 public class JSFunctionRef: JSObjectRef {
     @discardableResult
-    public func dynamicallyCall(withArguments arguments: [JSValueConvertible]) -> JSValue {
-        let result = arguments.withRawJSValues { rawValues -> RawJSValue in
+    func callAsFunction(this: JSObjectRef? = nil, args: [JSValueConvertible]) -> JSValue {
```
kateinoigakukun, line 5:
```suggestion
    public func callAsFunction(this: JSObjectRef? = nil, args: [JSValueConvertible]) -> JSValue {
```

Sources/JavaScriptKit/JSFunction.swift
```diff
@@ -1,45 +1,44 @@
...
+        new(args: args)
     }
 
-    public func new(_ arguments: JSValueConvertible...) -> JSObjectRef {
-        return arguments.withRawJSValues { rawValues in
+    // Guaranteed to return an object because either:
+    // a) the constructor explicitly returns an object, or
+    // b) the constructor returns nothing, which causes JS to return the `this` value, or
+    // c) the constructor returns undefined, null or a non-object, in which case JS also returns `this`.
+    public func new(args: [JSValueConvertible]) -> JSObjectRef {
```
kateinoigakukun, line 40:
```suggestion
    public func new(arguments: [JSValueConvertible]) -> JSObjectRef {
```

I think there shouldn't be abbreviations in public interface. 🙏

PR #30, opened 2020-08-04
Implement `RandomAccessCollection` on `JSArrayRef`

MaxDesiatov:
This allows using `Int` subscripts on `JSArrayRef` instances and makes the rest of the functions available from the relevant extensions. This may shadow some JavaScript functions with the same names (I don't remember any off the top of my head), but I hope that the added convenience is worth it.

Also `.gitignore` is amended to include the `.build` directory created by `make test`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 dist
 node_modules
 .DS_Store
-/.build
+.build
 /Packages
 /*.xcodeproj
 xcuserdata/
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
After having another look, `JSArrayRef` doesn't have dynamic member subscripts, so none of the JS properties or functions are exposed on it, no chance of collision then. If a user needs to access those they should just cast that object to `JSObjectRef` through the `.object` property of `JSValue` instead of `.array`.

PR #31, opened 2020-08-04
Add a helper method to copy an array of numbers to a JS TypedArray

j-f1:
Extracted from #26.


Sources/_CJavaScriptKit/include/_CJavaScriptKit.h
```diff
@@ -18,6 +18,20 @@ typedef enum __attribute__((enum_extensibility(closed))) {
...
+    JavaScriptTypedArrayKindUint16 = 3,
+    JavaScriptTypedArrayKindInt32 = 4,
+    JavaScriptTypedArrayKindUint32 = 5,
+    JavaScriptTypedArrayKindBigInt64 = 6,
+    JavaScriptTypedArrayKindBigUint64 = 7,
+    JavaScriptTypedArrayKindFloat32 = 8,
+    JavaScriptTypedArrayKindFloat64 = 9,
+} JavaScriptTypedArrayKind;
+
+
```
MaxDesiatov, line 34: Formatting consistency nitpick 🙂

```suggestion
```

j-f1:
I did a quick search, and it seems like a TypedArray is used as input to the `Blob` constructor. There’s also ways to construct TypedArrays out of the `ArrayBuffer` you get back from a `Blob`. The `TypedArray` stuff is separate from the main part of DOMKit, and there is no support for creating one from an `ArrayBuffer` yet, so we could push that feature to a future PR.

j-f1:
Also happy to write something that goes the other way now, though.

Runtime/src/index.ts
```diff
@@ -370,6 +396,16 @@ export class SwiftRuntime {
               const constructor = this.heap.referenceHeap(constructor_ref)
               return obj instanceof constructor
             },
+            swjs_create_typed_array: (
+                kind: JavaScriptTypedArrayKind,
+                elementsPtr: pointer, length: number,
+                result_obj: pointer
+            ) => {
+                const ArrayType: TypedArray = this.heap.referenceHeap(0)[JavaScriptTypedArrayKind[kind] + 'Array']
```
kateinoigakukun, line 404: Please don't abuse address `0` just to access global context.
It would be better to store global or window as a member of SwiftRuntime and access it directly 👍

Runtime/src/index.ts
```diff
@@ -370,6 +396,16 @@ export class SwiftRuntime {
...
+                const array = new ArrayType(memory().buffer, elementsPtr, length);
+                // Call `.slice()` to copy the memory
+                writeUint32(result_obj, this.heap.allocHeap(array.slice()));
```
kateinoigakukun, line 407: Why need to copy the object?

IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
```diff
@@ -334,3 +334,10 @@ ObjectRef_Lifetime: do {
 } catch {
     print(error)
 }
+
+TypedArray: do {
+    let numbers = [UInt8](0 ... 255)
+    let typedArray = JSObjectRef.createTypedArray(numbers)
+    try expectEqual(typedArray[12], .number(12))
+    try expectEqual(typedArray.toString!(), .string(numbers.map(String.init).joined(separator: ",")))
```
kateinoigakukun, line 342: Please add test cases for all types of typed array

Sources/JavaScriptKit/JSObject.swift
```diff
@@ -57,4 +57,67 @@ public class JSObjectRef: Equatable {
     public func jsValue() -> JSValue {
         .object(self)
     }
+
+    public class func createTypedArray<Type>(_ array: [Type]) -> JSObjectRef where Type: TypedArrayElement {
```
kateinoigakukun, line 61: Why don't you provide new type interface like `class JSTypedArray<Element: JSTypedArrayElement>`  instead of publishing low level operation.

Sources/JavaScriptKit/JSObject.swift
```diff
@@ -57,4 +57,67 @@ public class JSObjectRef: Equatable {
...
+        let type: JavaScriptTypedArrayKind
+        switch Type.self {
+        case is Int8.Type:
```
kateinoigakukun, line 64: Please add protocol method as requirement of TypedArrayElement to convert type into JavaScriptTypedArrayKind instead of switch-case types. It would be more exhaustive

Runtime/src/index.ts
j-f1: Because if we don't the Swift and JS arrays will share the memory which is not the point of the API.

Runtime/src/index.ts
j-f1: I created a `globalVariable` variable as a replacement. Is that a good name or should I go with something else?

j-f1:
I’ve created a `JSTypedArray` protocol that all the concrete types inherit from. Should I do it this way (matching the JS API) or just make `JSTypedArray` a generic class?

j-f1:
Another consideration: While working on #26, I changed the `JSTypedArray` class to wrap a regular `JSObjectRef` (like `JSArrayRef` does) instead of extending it. That makes use of a `JSBridgedClass` type which I could implement in a separate PR.

kateinoigakukun:
> I’ve created a JSTypedArray protocol that all the concrete types inherit from. Should I do it this way (matching the JS API) or just make JSTypedArray a generic class?


I prefer current generic class version. 👍

j-f1:
> I'd like to see the `JSBridgedClass` impl. Could you send another PR to review it?

It’s not 100% clean yet (there’s a few other things in the PR that are unrelated) but I added it in #26. ([see diff between it and this PR](https://github.com/j-f1/forked-JavaScriptKit/compare/typed-array...webidl?w=1))

There’s the `JSBridgedType` protocol which adds an `init?(from: JSValue)` initializer, which returns `nil` if the given value isn’t supported by the implementation. It extends `JSValueCodable` (aka `JSValueConvertible` & `JSValueConstructible`) so you can convert it back and forth from a `JSValue` using those APIs [maybe I should change it so that `JSValueConvertible` uses a `jsValue` read-only property and you do `SomeJSValueConstructible(from: JSValue)` instead of having a `static func construct` method?]

Then there’s the `JSBridgedClass` protocol which has a static `classRef` property pointing to the JS constructor for the class being represented and an instance `objectRef` property containing the `JSObjectRef` being wrapped. The `init?(from: JSValue)` implementation in `JSBridgedClass` ensures that the passed value is an object and is an instance of the `classRef` before calling `init(withCompatibleObject: JSObjectRef)` to delegate to the specific implementation.

> For example, if JavaScriptKit treats JSFunction as well as JSArrayRef, when casting a JSObjectRef into JSFunctionRef, JavaScript cast system needs to check that the object is a function using `obj instanceof Function`. But how can we get `Function` class? It can't.

I might be misunderstanding this but there is a `Function` property on the JS global object that all functions (including generator, async, and arrow functions) are `instanceof`. I do think it makes sense to treat functions differently since they have a different `typeof` though.

MaxDesiatov:
I'm thinking of generating some WebGLKit code, thankfully [an IDL file for it](https://github.com/KhronosGroup/WebGL/blob/master/specs/latest/1.0/webgl.idl) is available. Using typed arrays is pretty much a requirement for WebGL, so I'd be very happy for this PR to move forward. Is there any way I can help here?

j-f1:
The main sticking point is on the bridging API. I think the one in #26 can be extracted out to another PR, if @kateinoigakukun likes it.

kateinoigakukun:
This is my change proposal https://github.com/j-f1/forked-JavaScriptKit/pull/1

If you accept my proposal, I can immediately merge your PR swiftwasm#31

kateinoigakukun (approved):

PR #38, opened 2020-08-27
Fix invalid array termination for null and undefined

kateinoigakukun:


MaxDesiatov (approved):
Seems legit 👍

j-f1:
Not entirely sure how this should be handled, but JS Arrays support “holes”, where `[1,,3]` creates an array without anything at index `1`. From the outside, that index appears to contain `undefined`, but array methods like `map` and `forEach` skip over those indices while iterating, and `flat` removes them, changing the indices of everything that comes after the hole. You can tell the difference between a hole and an `undefined` value with either `index in array` or `array.hasOwnProperty(index)`.

I think it’s perfectly fine to simply say that holes are not handled properly since they’re quite rare in practice but I wanted to make sure this language quirk wasn’t missed.

MaxDesiatov:
Do present test cases handle this quirk adequately or is there anything else you'd like to be covered in the test suite?

j-f1:
I don’t believe there are any tests that involve array holes in the current test suite so if we want to support it new tests would need to be added.

PR #40, opened 2020-08-28
Refine public API

kateinoigakukun:
Drop Ref suffix


Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -1,8 +1,8 @@
 import _CJavaScriptKit
 
-public class JSFunctionRef: JSObjectRef {
+public class JSFunctionRef: JSObject {
```
j-f1, line 3: Should this be `JSFunction`?
```suggestion
public class JSFunction: JSObject {
```

kateinoigakukun: Oops, I missed that. Thanks

PR #41, opened 2020-08-28
Support Holes in Array

### LINKED ISSUE #39: Support for JS Arrays “holes”, including the test suite
Not entirely sure how this should be handled, but JS Arrays support “holes”, where `[1,,3]` creates an array without anything at index `1`. From the outside, that index appears to contain `undefined`, but array methods like `map` and `forEach` skip over those indices while iterating, and `flat` removes them, changing the indices of everything that comes after the hole. You can tell the difference between a hole and an `undefined` value with either `index in array` or `array.hasOwnProperty(index)`.

I think it’s perfectly fine to simply say that holes are not handled properly since they’re quite rare in practice but I wanted to make sure this language quirk wasn’t missed.

_Originally posted by @j-f1 in https://github.com/swiftwasm/JavaScriptKit/pull/38#issuecomment-682083583_

kateinoigakukun:
Closes #39

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSArray.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSArray.swift
@@ -47,7 +47,7 @@ extension JSArray: RandomAccessCollection {
 
     public var startIndex: Int { 0 }
 
-    public var endIndex: Int { ref.length.number.map(Int.init) ?? 0 }
+    public var endIndex: Int { count }
 
     /// The number of elements in that array including empty hole.
     /// Note that `length` respects JavaScript's `Array.prototype.length`
@@ -73,9 +73,9 @@ extension JSArray: RandomAccessCollection {
     }
 }
 
+private let alwaysTrue = JSClosure { _ in .boolean(true) }
 private func getObjectValuesLength(_ object: JSObject) -> Int {
-    let objectClass = JSObject.global.Object.function!
-    let values = objectClass.values!(object).object!
+    let values = object.filter!(alwaysTrue).object!
     return Int(values.length.number!)
 }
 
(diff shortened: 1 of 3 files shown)


Sources/JavaScriptKit/BasicObjects/JSArray.swift
```diff
@@ -44,6 +48,35 @@ extension JSArray: RandomAccessCollection {
...
+    /// Note that `count` syncs with the number that `Iterator` can iterate.
+    /// See also: `JSArray.length`
+    public var count: Int {
+        return getObjectValuesLength(ref)
+    }
+}
+
+private func getObjectValuesLength(_ object: JSObject) -> Int {
+    let objectClass = JSObject.global.Object.function!
+    let values = objectClass.values!(object).object!
```
j-f1, line 78: This won't work in another edge case — when there are additional properties added to the array (like doing `array.foo = 'bar'`).

According to [this post](https://2ality.com/2015/09/holes-arrays-es6.html), running `array.filter(() => true)` removes the holes so that could be used to get a better count.

Sources/JavaScriptKit/BasicObjects/JSArray.swift
```diff
@@ -43,7 +47,36 @@ extension JSArray: RandomAccessCollection {
 
     public var startIndex: Int { 0 }
 
-    public var endIndex: Int { ref.length.number.map(Int.init) ?? 0 }
+    public var endIndex: Int { count }
```
j-f1, line 50: Shouldn't this be `length - 1` since the subscript notation does count holes? Or is this important to using the iterator?

kateinoigakukun: Sorry, I accidentally changed that.

PR #45, opened 2020-09-13
Add JSDate implementation with tests

MaxDesiatov:
This doesn't 100% match the JS API, for example `getMonth`/`setMonth` etc accessor methods are converted to properties, but the rest of it matches in the naming. I didn't expose API parts [that MDN marks as non-standard](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) or not consistent across browsers and JS implementations.

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSDate.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSDate.swift
@@ -1,0 +1,203 @@
+public final class JSDate {
+    private static let constructor = JSObject.global.Date.function!
+    private let ref: JSObject
+
+    public init(millisecondsSinceEpoch: Int? = nil) {
+        if let milliseconds = millisecondsSinceEpoch {
+            ref = Self.constructor.new(milliseconds)
+        } else {
+            ref = Self.constructor.new()
+        }
+    }
+
+    /** According to the standard, `monthIndex` is zero-indexed, where `11` is December. `day` 
+    represents a day of the month starting at `1`.
+    */
+    public init(
+        year: Int,
+        monthIndex: Int,
+        day: Int = 1,
+        hours: Int = 0,
+        minutes: Int = 0,
+        seconds: Int = 0,
+        milliseconds: Int = 0
+    ) {
+        ref = Self.constructor.new(year, monthIndex, day, hours, minutes, seconds, milliseconds)
+    }
+
+    /// Year of this date in local time zone.
+    public var fullYear: Int {
+        get {
+            Int(ref.getFullYear.function!().number!)
+        }
+        set {
+            ref.setFullYear.function!(newValue)
+        }
+    }
+
+    /// Month of this date in `0–11` range in local time zone
+    public var month: Int {
+        get {
+            Int(ref.getMonth.function!().number!)
+        }
+        set {
+            ref.setMonth.function!(newValue)
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


Sources/JavaScriptKit/BasicObjects/JSDate.swift
```diff
@@ -0,0 +1,203 @@
...
+        seconds: Int = 0,
+        milliseconds: Int = 0
+    ) {
+        ref = Self.constructor.new(year, monthIndex, day, hours, minutes, seconds, milliseconds)
+    }
+
+    /// Year of this date in local time zone.
+    public var fullYear: Int {
+        get {
+            Int(ref.getFullYear.function!().number!)
```
j-f1, line 31: I think this can just be `Int(ref.getFullYear!().number!)`. Same for the other function calls

Sources/JavaScriptKit/BasicObjects/JSDate.swift
```diff
@@ -28,176 +28,183 @@ public final class JSDate {
...
-    /// Number of seconds since epoch ignoring leap seconds
-    public func now() -> Int {
-        Int(ref.now.function!().number!)
+    /// Number of seconds since midnight 01 January 1970 UTC to the present moment ignoring leap 
+    /// seconds
+    public static func now() -> Double {
+        constructor.now!().number!
+    }
+
+    /// Number of seconds since midnight 01 January 1970 UTC to the present moment ignoring leap
```
j-f1, line 205:
```suggestion
    /// Number of milliseconds since midnight 01 January 1970 UTC to the present moment ignoring leap 
```

Sources/JavaScriptKit/BasicObjects/JSDate.swift
```diff
@@ -28,176 +28,183 @@ public final class JSDate {
...
 
     public func toUTCString() -> String {
-        ref.toUTCString.function!().string!
+        ref.toUTCString!().string!
     }
 
-    /// Number of seconds since epoch ignoring leap seconds
-    public func now() -> Int {
-        Int(ref.now.function!().number!)
+    /// Number of seconds since midnight 01 January 1970 UTC to the present moment ignoring leap
```
j-f1, line 199:
```suggestion
    /// Number of milliseconds since midnight 01 January 1970 UTC to the present moment ignoring leap 
```

Sources/JavaScriptKit/BasicObjects/JSDate.swift
MaxDesiatov: great catch 👍

Sources/JavaScriptKit/BasicObjects/JSDate.swift
```diff
@@ -0,0 +1,220 @@
...
+        ref.toUTCString!().string!
+    }
+
+    /// Number of milliseconds since midnight 01 January 1970 UTC to the present moment ignoring leap 
+    /// seconds
+    public static func now() -> Double {
+        constructor.now!().number!
+    }
+
+    /// Number of milliseconds since midnight 01 January 1970 UTC to the present moment ignoring leap
```
MaxDesiatov, line 205:
```suggestion
    /// Number of milliseconds since midnight 01 January 1970 UTC to the given date ignoring leap 
```

j-f1 (approved):
LGTM!

kateinoigakukun (approved):

PR #46, opened 2020-09-13
Add JSTimer implementation with tests

MaxDesiatov:
Depends on #45. This is also a prerequisite for a future `JSPromise` PR with tests.

I intentionally didn't match the JS API in this PR, as a special care is needed to hold a reference to the timer closure and to call `.release()` on it. Here, a user is supposed to hold a reference to a `JSTimer` instance for it to stay valid. The API is also intentionally simple, the timer is started right away, and the only way to invalidate the timer is to bring its reference count to zero.

If you see a better way to manage closures passed to `setTimeout` and `setInterval`, I'd be happy to consider that.

Also, Node.js and browser APIs are slightly different. `setTimeout`/`setInterval` return an object in Node.js, while browsers return a number. Fortunately, `clearTimeout` and `clearInterval` take corresponding types as their arguments, and we can store either as `JSValue`, so we can treat both cases uniformly.

### DIFF
--- a/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
+++ b/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
@@ -434,3 +434,33 @@ try test("Date") {
 
     try expectEqual(date3 < date1, true)
 }
+
+// make the timers global to prevent early deallocation
+var timeout: JSTimer?
+var interval: JSTimer?
+
+try test("Timer") {
+    let start = JSDate().valueOf()
+    let timeoutMilliseconds = 5.0
+
+    timeout = JSTimer(millisecondsDelay: timeoutMilliseconds, isRepeating: false) {
+        // verify that at least `timeoutMilliseconds` passed since the `timeout` timer started
+        try! expectEqual(start + timeoutMilliseconds <= JSDate().valueOf(), true)
+    }
+
+    var count = 0.0
+    let maxCount = 5.0
+    interval = JSTimer(millisecondsDelay: 5, isRepeating: true) {
+        // verify that at least `timeoutMilliseconds * count` passed since the `timeout` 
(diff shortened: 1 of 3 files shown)


j-f1:
- I think it would make sense to use the standard `TimeInterval` type here and convert it to the appropriate millisecond value to better match other Swift APIs
- the API is pretty verbose. What do you think of this?

```swift
public final class JSTimer {
  public init(_ after: TimeInterval, _ callback: @escaping () -> ())
  public static repeating(_ interval: TimeInterval, _ callback: @escaping () -> ())
}
```

kateinoigakukun:
@j-f1

> I think it would make sense to use the standard TimeInterval type here and convert it to the appropriate millisecond value to better match other Swift APIs

It's reasonable. But unfortunately, `TimeInterval` is a type defined in Foundation and I don't want JavaScriptKit to depend on Foundation.

MaxDesiatov:
@j-f1 
>     * the API is pretty verbose. What do you think of this?
> 
> ```swift
> public final class JSTimer {
>   public init(_ after: TimeInterval, _ callback: @escaping () -> ())
>   public static repeating(_ interval: TimeInterval, _ callback: @escaping () -> ())
> }
> ```

I'm not sure this would work. `repeating` still has to delegate to some initializer, but there isn't one in this case that would call `setInterval` if I assume that the only one calls `setTimeout`. Then it would have to look something like:

```swift
public final class JSTimer {
  public init(millisecondsDelay: Double, isRepeating: Bool = false, callback: @escaping () -> ())
  public static func repeating(millisecondsDelay: Double, callback: @escaping () -> ())
}
```

Maybe we could just add a default `false` value on `isRepeating` without adding `static func repeating`? WDYT?

kateinoigakukun (approved):

PR #47, opened 2020-09-13
Add JSError with tests, add JSObject.description

MaxDesiatov:
`JSError` will be used in a future PR with a `JSPromise` implementation to make a statically typed `catch` method available on it.

While there's no straightforward way to throw `JSError` from Swift code right now, I think this type might become useful in some other cases too in the future. Maybe we could find a way to catch an error from an arbitrary JavaScript function?

`JSObject` and `JSError` are also `CustomStringConvertible`, as `toString()` is available on all JavaScript objects according to the standard.


Sources/JavaScriptKit/BasicObjects/JSError.swift
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    public var message: String {
+        ref.message.string!
+    }
+
+    public var name: String {
+        ref.name.string!
+    }
+}
```
j-f1, line 16: Although it’s not standard, all engines that I know of implement `error.stack` to provide a stack trace as a string. It might be useful to access it.

kateinoigakukun (approved):

PR #48, opened 2020-09-14
Add JSError.stack, add Error conformance

MaxDesiatov:
While the `stack` property is non-standard, it's [supported in all popular browser engines](https://caniuse.com/mdn-javascript_builtins_error_stack) and Node.js.

Unfortunately, because `stack` value will be different on every machine that executes it, I'm not sure how to write a good test for it, so it currently doesn't have any tests.

Also, `JSError` now conforms to `Error`. The main reasoning is that the `Publisher` protocol in Combine requires `Error` conformance on its `Failure` type. I think in the future it would make sense to make `JSPromise` compatible with Combine, so it would be great if one could propagate errors produced by `JSPromise` to other publishers/subscribers.

Again, `JSPromise` will be implemented in a separate PR.

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSError.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSError.swift
@@ -1,4 +1,4 @@
-public final class JSError {
+public final class JSError: Error {
     private let ref: JSObject
     private static let constructor = JSObject.global.Error.function!
 
@@ -12,6 +12,10 @@ public final class JSError {
 
     public var name: String {
         ref.name.string!
+    }
+
+    public var stack: String {
+        ref.stack.string!
     }
 }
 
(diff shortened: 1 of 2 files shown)


j-f1:
You could test that the stack is a non-empty string.

MaxDesiatov:
Great point, I've added the check to the test.

j-f1 (approved):

kateinoigakukun (approved):

Sources/JavaScriptKit/BasicObjects/JSError.swift
```diff
@@ -13,6 +13,10 @@ public final class JSError {
     public var name: String {
         ref.name.string!
     }
+
+    public var stack: String {
```
kateinoigakukun, line 17: I think it would be more safe if non-standard field is exposed as `String?`. For example, a new JS engine may not implement this feature.

PR #49, opened 2020-09-14
Generate and publish documentation with `swift-doc`

MaxDesiatov:
You can see the deployment at https://swiftwasm.github.io/JavaScriptKit/.

I'll update the trigger to deploy only on pushes to `master` if you think this is something we should proceed with.

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -1,6 +1,8 @@
 name: Documentation
 
-on: [push]
+on: 
+  push:
+    branches: [master]
 
 jobs:
   swift-doc:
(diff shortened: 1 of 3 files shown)


j-f1:
Looks like there’s some cleanup necessary in the “functions” category — all of those should be private imo. It also doesn’t properly handle standard Swift types — is there a way to link those to developer.apple.com?

MaxDesiatov:
I'll clean things up, but cross-module references unfortunately aren't supported in `swift-doc` yet 😔

MaxDesiatov:
Everything seems to work when `@_cdecl` functions are made internal, so they are no longer displayed in the docs now.

kateinoigakukun (approved):
Nice!

PR #50, opened 2020-09-14
Rename `ref` to `jsObject` on JSDate for consistency with JSError

MaxDesiatov:
I think that `jsObject` name is more representative than plain `ref`, and as Jed correctly points out, it should be public after all for users to access it if necessary.

Also, more doc comments are added to `JSDate` with existing ones refined.


j-f1:
I think this is a useful thing to have public so users can access the raw JS value if necessary.

kateinoigakukun (approved):

j-f1 (approved):

PR #57, opened 2020-09-16
Add doc comments for public APIs (Part 2)

kateinoigakukun:

PR #58, opened 2020-09-16
Remove JavaScriptTypedArrayKind

### LINKED ISSUE #52: TypedArray improvement?
Currently, creating a TypedArray using `swjs_create_typed_array` requires passing an enumerated value, `JavaScriptTypedArrayKind`, to determine which kind of typed array should be created. I don’t remember exactly why I did this — would it make more sense to just pass a reference to the constructor function for the particular typed array we want?

j-f1:
Fixes #52.

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -117,7 +117,7 @@ class SwiftRuntimeHeap {
 export class SwiftRuntime {
     private instance: WebAssembly.Instance | null;
     private heap: SwiftRuntimeHeap
-    private version: number = 600
+    private version: number = 610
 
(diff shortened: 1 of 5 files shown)


j-f1:
Reading through #57 reminded me that I did not bump the library version in #31. Is this an issue? Also, what’s the standard for how much to increment the version?

kateinoigakukun:
> I did not bump the library version in #31. Is this an issue?

This is not a big issue because I bumped the version before releasing 0.6.0.

> Also, what’s the standard for how much to increment the version?

At least it's ok if it's greater than the old value.

PR #59, opened 2020-09-16
Make JSValue ExpressibleByNilLiteral

j-f1:
I have `nil` produce `.null` now but should it produce `.undefined` instead maybe? This is the last thing needed to get DOMKit building without `AnyJSValueConvertible`.


MaxDesiatov (approved):
I think `.null` is appropriate here, `.undefined` corresponds more closely to `()`/`Void` in my opinion.

kateinoigakukun (approved):

Issue #61, opened 2020-09-16
JavaScriptKit as a "system" library?


MaxDesiatov:
It would be great if we could provide shims for Dispatch and Foundation APIs that rely on JS hosts directly instead of WASI. The problem is that Dispatch and Foundation are "system" modules shipped with the SwiftWasm SDK, while JavaScriptKit is consumed by users via SwiftPM.

I wonder if we could consider promoting JavaScriptKit to a "system" library, I guess Swift for Tensorflow does a similar thing with [PythonKit](https://github.com/pvieito/PythonKit) already. This will obviously require our API to become much more stable, but I hope it would be worth it.

WDYT?

kateinoigakukun:
I think that would be a good option. However, I think it's too early to ship this in the toolchain due to compatibility issues.

kateinoigakukun:
Closing as this will unlikely happen

PR #62, opened 2020-09-16
Add a generic `JSPromise` implementation

MaxDesiatov:
This provides three overloads each for `then` and `catch`, and I'm not sure if that's good for the type-checker performance and usability. I was thinking of naming the promise-returning callback version `flatMap`, but ultimately decided against it.

`then` overload with two callbacks is not available, because it's impossible to unify success and failure types from both callbacks in a single returned promise without type erasure. I think users should just chain `then` and `catch` in those cases so that type erasure is avoided.

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
@@ -1,0 +1,132 @@
+/** A wrapper around [the JavaScript `Promise` class](https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+that exposes its functions in a type-safe and Swifty way. The `JSPromise` API is generic over both
+`Success` and `Failure` types, which improves compatibility with other statically-typed APIs such
+as Combine. If you don't know the exact type of your `Success` value, you should use `JSValue`, e.g.
+`JSPromise<JSValue, JSError>`. In the rare case, where you can't guarantee that the error thrown
+is of actual JavaScript `Error` type, you should use `JSPromise<JSValue, JSValue>`.
+*/
+public final class JSPromise<Success, Failure>: JSValueConvertible, JSValueConstructible
+where Success: JSValueConstructible, Failure: JSValueConstructible {
+    /// The underlying JavaScript `Promise` object.
+    public let jsObject: JSObject
+
+    private var callbacks = [JSClosure]()
+
+    /// The underlying JavaScript `Promise` object wrapped as `JSValue`.
+    public func jsValue() -> JSValue {
+        .object(jsObject)
+    }
+
+    /// This private initializer assumes that the passed object is a JavaScript `Promise`
+    private init(unsafe object: JSObject) {
+        self.jsObject = object
+    }
+
+    /** Creates a new `JSPromise` instance from a given JavaScript `Promise` object. If `jsObject`
+    is not an instance of JavaScript `Promise`, this initializer will return `nil`.
+    */
+    public init?(_ jsObject: JSObject) {
+        guard jsObject.isInstanceOf(JSObject.global.Promise.function!) else { return nil }
+        self.jsObject = jsObject
+    }
+
+    /** Creates a new `JSPromise` instance from a given JavaScript `Promise` object. If `value`
+    is not an object and is not an instance of JavaScript `Promise`, this function will 
+    return `nil`.
+    */
+    public static func construct(from value: JSValue) -> Self? {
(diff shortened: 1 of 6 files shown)


Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /** Schedules the `failure` closure to be invoked on rejected completion of `self`.
+    */
+    public func `catch`(failure: @escaping (Failure) -> ()) {
+        let closure = JSClosure {
+            failure(Failure.construct(from: $0[0])!)
+        }
+        callbacks.append(closure)
+        jsObject.then.function!(JSValue.undefined, closure)
```
j-f1, line 91: Is there a reason you’re not using `jsObject.catch!(closure)` here?

j-f1: Also the difference between `foo.bar.function!(...)` and `foo.bar!(...)` is that the latter preserves the automatic `this`-binding JS does, which is important for object methods that are likely not already bound to their object.

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -0,0 +1,136 @@
...
+    public static func construct(from value: JSValue) -> Self? {
+        guard case let .object(jsObject) = value else { return nil }
+        return Self.init(jsObject)
+    }
+
+    /** Schedules the `success` closure to be invoked on sucessful completion of `self`.
+    */
+    public func then(success: @escaping (Success) -> ()) {
+        let closure = JSClosure {
+            success(Success.construct(from: $0[0])!)
```
kateinoigakukun, line 50: If fail to construct, we should throw error or report it with a detailed message. Force unwrapping fatal messages is not quite useful to see the error reason for JavaScriptKit users.

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -0,0 +1,136 @@
...
+`Success` and `Failure` types, which improves compatibility with other statically-typed APIs such
+as Combine. If you don't know the exact type of your `Success` value, you should use `JSValue`, e.g.
+`JSPromise<JSValue, JSError>`. In the rare case, where you can't guarantee that the error thrown
+is of actual JavaScript `Error` type, you should use `JSPromise<JSValue, JSValue>`.
+
+This doesn't 100% match the JavaScript API, as `then` overload with two callbacks is not available.
+It's impossible to unify success and failure types from both callbacks in a single returned promise
+without type erasure. You should chain `then` and `catch` in those cases to avoid type erasure.
+*/
+public final class JSPromise<Success, Failure>: JSValueConvertible, JSValueConstructible
```
kateinoigakukun, line 12: Please note that this class must have the same lifetime with the actual `Promise` object in JS environment because callback handlers will be deallocated when `JSPromise.deinit`.

If the actual `Promise` object in JS environment lives longer than this `JSPromise`, it may call deallocated `JSClosure`.

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
MaxDesiatov: > Is there a reason you’re not using `jsObject.catch!(closure)` here?

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch):

> It behaves the same as calling `Promise.prototype.then(undefined, onRejected)` (in fact, calling `obj.catch(onRejected)` internally calls `obj.then(undefined, onRejected))`.

j-f1:
I just had an idea: what if the JSPromise class exposed an `ObservableObject`? We could have some code on the JS side that would attach itself to the promise and tell Swift to update either the `value` or `error` property without having to worry about managing Swift closures.

MaxDesiatov:
That will require a dependency on OpenCombine unfortunately, and it would be best to keep JavaScriptKit the lowest level library in the stack if we ever hope to achieve https://github.com/swiftwasm/JavaScriptKit/issues/61. What about a separate OpenCombineDOM library that allows any `JSPromise` to become a publisher, and then exposes `fetch` and timer publishers too? Yes, this multiplies the number of libraries, but the JS ecosystem is fragmented (read "modular") anyway, and sadly we'd have to reflect that to allow users to mix and match what they intend to use in their apps.

j-f1 (approved):

kateinoigakukun (approved):

PR #63, opened 2020-09-18
Implement JSString to reduce bridging overhead

kateinoigakukun:
Before

```
node bin/benchmark-tests.js
Running 'Serialization/Write JavaScript number directly' ...
done 123.83756804466248 ms
Running 'Serialization/Write JavaScript string directly' ...
done 129.04579210281372 ms
Running 'Serialization/Swift Int to JavaScript' ...
done 3677.7331490516663 ms
Running 'Serialization/Swift String to JavaScript' ...
done 5490.698668003082 ms
Running 'Object heap/Increment and decrement RC' ...
done 3000.8530011177063 ms
```

After

```
node bin/benchmark-tests.js
Running 'Serialization/Write JavaScript number directly' ...
done 130.88212418556213 ms
Running 'Serialization/Write JavaScript string directly' ...
done 137.1980230808258 ms
Running 'Serialization/Swift Int to JavaScript' ...
done 3526.049220085144 ms
Running 'Serialization/Swift String to JavaScript' ...
done 3598.435418844223 ms
Running 'Object heap/Increment and decrement RC' ...
done 2782.3261749744415 ms
```

PR #64, opened 2020-09-18
Remove payload2 from value bridging interface

kateinoigakukun:
Depends on #63

### DIFF
--- a/Sources/JavaScriptKit/JSValueConvertible.swift
+++ b/Sources/JavaScriptKit/JSValueConvertible.swift
@@ -183,7 +183,7 @@ extension RawJSValue: JSValueConvertible {
         case .boolean:
             return .boolean(payload1 != 0)
         case .number:
-            return .number(payload3)
+            return .number(payload2)
         case .string:
             return .string(JSString(jsRef: payload1))
         case .object:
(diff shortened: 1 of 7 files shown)


MaxDesiatov (approved):
This seems legit as the tests pass, but could we have a doc comment somewhere that clarifies the purpose, the number, and the types of payloads? Is was going to provide a write up on JavaScriptKit internals, would be very interested to know more in details, if possible 🙏

Sources/JavaScriptKit/JSValueConvertible.swift
```diff
@@ -194,8 +194,6 @@ extension RawJSValue: JSValueConvertible {
             return .undefined
         case .function:
             return .function(JSFunction(id: UInt32(payload1)))
-        default:
-            fatalError("unreachable")
```
j-f1, line 198: What happens if this is passed an invalid `kind`?

kateinoigakukun: The swift runtime will crash with error message that says the given enum value has invalid raw value.

j-f1 (approved):

MaxDesiatov:
Thanks a lot for the diagrams! Happy for this to be merged, and maybe if you folks manage to review #62, it would be a good 0.7.0 release? 😉

PR #80, opened 2020-09-25
Run 4 perf tests instead of 2

j-f1:
It looks like the perf test script is running in ~2m before this change, which leaves plenty of time to gather more samples to hopefully smooth out the results so perf changes can be better caught.

I changed it to run the perf tests four times, with 2 in parallel at once. The constants can be easily changed if we find it’s still not stable enough.


MaxDesiatov (approved):
0% here is good. I guess we'll need to see it running on more PRs to understand if the variance is really lower with 4 runs instead of 2.

PR #81, opened 2020-09-26
Fix runtime files location in `package.json`

### LINKED ISSUE #79: 0.7.0 unavailable on NPM
I know that new versions take time to show on the NPM frontend, but usually installing a newly published version is possible almost immediately. 0.7.0 currently isn't available when installing with NPM, is this because 0.7.0 hasn't been published yet, or is this because of NPM caches not being updated?

MaxDesiatov:
Resolves #79.

This requires tagging a new release and republishing on NPM to fully fix the issue.


j-f1 (approved):
Yikes, thanks for fixing this!

PR #84, opened 2020-09-26
Automatically publish to npm

j-f1:
I copied the code from GitHub’s example.

This should activate whenever a new release is tagged.

@kateinoigakukun needs to upload an npm token to the secrets section of the repo for this to work.

Let’s give this a try for v0.7.1.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -8,18 +8,7 @@ on:
     types: [created]
 
 jobs:
-  build:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
-        with:
-          node-version: 12
-      - run: npm ci
-      - run: npm test
-
   publish-npm:
-    needs: build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -1,0 +1,33 @@
+# This workflow will run tests using node and then publish a package to npm when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Publish to npm
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci
+      - run: npm test
```
MaxDesiatov, line 19: We don't have `ci` or `test` scripts defined in `package.json`, so these steps will probably fail.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,33 @@
...
+  publish-npm:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+          registry-url: https://registry.npmjs.org/
+      - run: npm ci
```
MaxDesiatov, line 30: Same here, wouldn't it require a `ci` script to be defined?

kateinoigakukun:
Could you test this script on another test repo?

.github/workflows/npm-publish.yml
j-f1: `npm ci` is like `npm install` but it never modifies `package.json`.

.github/workflows/npm-publish.yml
j-f1: Good point! I’ve removed the testing step.

MaxDesiatov:
I've verified [it works in my `gatsby-remark-codemirror` package](https://github.com/MaxDesiatov/gatsby-remark-codemirror/runs/1172488500?check_suite_focus=true). Good to go, you just need to add the `npm_token` secret, and make sure your OTP is authorization-only if you have it enabled.

kateinoigakukun:
@MaxDesiatov @j-f1 Could you tell me your npm account name? I'll invite you as owner.

MaxDesiatov:
I'm `maxd` on npm 🙂

MaxDesiatov (approved):
👍

kateinoigakukun (approved):

j-f1:
I’m  `j-f` on npm.

PR #85, opened 2020-09-28
Add missing .npmignore, bump version to 0.7.2

MaxDesiatov:
In 0.7.1 the runtime files are still not available through npm. [The reason was this](https://docs.npmjs.com/files/package.json#files):

>If there is a `.gitignore` file, and `.npmignore` is missing, `.gitignore`’s contents will be used instead.

The `Runtime` directory had `.gitignore` that mentioned the `lib` subdirectory with transpiled files. This made sense, but absence of `.npmignore` excluded these files from uploading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.7.2 (28 September 2020)
+
+This is a bugfix release that resolves an issue with the JavaScript runtime being unavailable when installed via NPM.
+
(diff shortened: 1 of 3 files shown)


j-f1:
The files are now present! https://unpkg.com/browse/javascript-kit-swift@0.7.2/Runtime/lib/

MaxDesiatov:
Awesome, thanks for tagging the release!

PR #86, opened 2020-09-29
Make `JSError` conform to `JSBridgedClass`

MaxDesiatov:
The lack of this conformance makes it impossible to attach `catch` callbacks to `JSPromise` when `Failure: JSError`, as `catch` methods have `Failure: JSValueConstructible` constraints.

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSError.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSError.swift
@@ -43,3 +43,10 @@ extension JSError: CustomStringConvertible {
     /// The textual representation of this error.
     public var description: String { jsObject.description }
 }
+
+extension JSError: JSValueConstructible {
+  public static func construct(from value: JSValue) -> JSError? {
+    guard let object = value.object else { return nil }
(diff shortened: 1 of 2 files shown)


j-f1:
Does it make sense to make JSError a bridged class to the JavaScript Error class?￼￼

kateinoigakukun (approved):

Sources/JavaScriptKit/BasicObjects/JSError.swift
```diff
@@ -43,3 +43,10 @@ extension JSError: CustomStringConvertible {
     /// The textual representation of this error.
     public var description: String { jsObject.description }
 }
+
+extension JSError: JSValueConstructible {
+  public static func construct(from value: JSValue) -> JSError? {
+    guard let object = value.object else { return nil }
+    return JSError(unsafelyWrapping: object)
```
kateinoigakukun, line 50: I think we need to guard non-Error object using isInstanceOf

j-f1: If you use `JSBridgedClass` it will handle that for you.

Issue #6, opened 2020-05-05
Build fails with the unsafe flags error


MaxDesiatov:
I have a very trivial app [here](https://github.com/MaxDesiatov/carton/tree/master/TestApp) that depends on JavaScriptKit, but it fails to build with this error:

```
% swift build --triple wasm32-unknown-wasi                                     
error: the target 'JavaScriptKit' in product 'JavaScriptKit' contains unsafe build flags
% swiftenv version
wasm-DEVELOPMENT-SNAPSHOT-2020-05-02-a (set by /Users/maxd/Documents/carton/TestApp/.swift-version)
% which swift
/Users/maxd/.swiftenv/shims/swift
% swift --version
Swift version 5.3-dev (LLVM 11086900c5, Swift 52a06ad8f3)
Target: x86_64-apple-darwin19.4.0
```

kateinoigakukun:
Yes, you can't use a tag to specify a fixed revision due to unsafe flags.
So you need to specify a commit hash to fix revision like [here](https://github.com/swiftwasm/swift-webpack-template/blob/master/Package.swift#L12)

But in the long run, we need to support these flags safely in SwiftPM.

MaxDesiatov:
I think this becomes a higher priority as it prevents any projects that depend on JavaScriptKit from being built for any platform other than non-WASI. For example, it's currently blocking https://github.com/TokamakUI/Tokamak/pull/276.

The big problem is that it may require changes to the package manifest API, such as new safe linker settings. Any new settings won't be available on non-WASI platforms until the next version of Swift is released. If we add new linker settings, they definitely need to land upstream before the next version is branched off, but I think we also need to find some workaround in the meantime.

MaxDesiatov:
Here's a possible solution that I proposed in the `#webassembly` channel of SwiftPM Slack:

1. For SwiftWasm 5.3 we could hardcode the flags and apply them in our forked SwiftPM if target/product name is JavaScriptKit. We then remove unsafe flags from `Package.swift` in JavaScriptKit as our forked SwiftPM would know how to build it anyway. We already [have an ugly hardcoded check](https://github.com/swiftwasm/swift-package-manager/blob/swiftwasm-release/5.3/Sources/PackageGraph/PackageGraphLoader.swift#L544) in our fork to allow unsafe flags in JavaScriptKit anyway.
[...]

PR #91, opened 2020-10-02
Remove all unsafe linker flags from Package.swift

kateinoigakukun:
Need toolchain including https://github.com/swiftwasm/llvm-project/commit/20c933e5d1d93c86a6f5b8fd6b1234eeef71d011 

Resolve https://github.com/swiftwasm/JavaScriptKit/issues/6


.github/workflows/perf.yml
```diff
@@ -16,6 +16,7 @@ jobs:
           export SWIFTENV_ROOT="$HOME/.swiftenv"
           export PATH="$SWIFTENV_ROOT/bin:$PATH"
           eval "$(swiftenv init -)"
+          ./scripts/install-toolchain.sh
           swiftenv install $TOOLCHAIN_DOWNLOAD
```
kateinoigakukun, line 20: We need to remove swiftenv install line after merging this PR because older version toolchain is necessary to compare with base branch

kateinoigakukun:
The reason for the performance improvement is that we started to build toolchain without assertion.

j-f1:
Is that expected? I’m pretty sure the benchmarks only measure runtime performance, not built time. Or am I missing something about what “building without assertions” means?

kateinoigakukun:
@j-f1 It's an expected result. Because stdlib binary is built without any `assert(condition)`, the runtime performance is improved.

MaxDesiatov:
Do you know if upstream distributions by the way? I thought that we disabled assertions temporarily until https://github.com/swiftwasm/swift/issues/1823 is fixed.

kateinoigakukun:
@MaxDesiatov IMO, we should disable assertion for distribution toolchains even https://github.com/swiftwasm/swift/issues/1823 will be fixed. I'm not sure which build-preset is used for upstream toolchain distribution, but they use no-assertion for release version toolchain in general.

kateinoigakukun:
@MaxDesiatov @j-f1 Can we merge this PR or do I need more detail description?

MaxDesiatov (approved):
Great stuff, I look forward to building Tokamak on Linux or macOS 🙂

j-f1 (approved):

PR #96, opened 2020-10-07
Mention the `install-toolchain.sh` script in `README.md`

MaxDesiatov:
The toolchain snapshot suggested in `README.md` is outdated. I think we should recommend using the script to install it automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,10 @@ Swift framework to interact with JavaScript through WebAssembly.
 
 ## Requirements
 
-This library only supports [`swiftwasm/swift`](https://github.com/swiftwasm/swift) distribution toolchain. Please install Swift for WebAssembly toolchain from [Release Page](https://github.com/swiftwasm/swift/releases)
-
-The toolchains can be installed via [`swiftenv`](https://github.com/kylef/swiftenv) like official nightly toolchain.
-
-e.g.
+This library only supports [`swiftwasm/swift`](https://github.com/swiftwasm/swift) distribution toolchain. The toolchains can be installed via [`swiftenv`](https://github.com/kylef/swiftenv) like official nightly toolchain. You can use the `install-toolchain.sh` helper script that does this for you:
 
 ```sh
-
-$ swiftenv install https://github.com/swiftwasm/swift/releases/download/swift-wasm-5.3-SNAPSHOT-2020-08-10-a/swift-wasm-5.3-SNAPSHOT-2020-08-10-a-osx.tar.gz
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Another option would be to suggest `carton` as the default build tool for end users. WDYT?

carson-katri:
I suppose both could be mentioned, then they can choose what fits their needs best?

kateinoigakukun:
I think we need to provide two section, the first is for users of JavaScriptKit and the second is for contributors of JavaScriptKit.
IMO, we should keep swiftenv installation section for users of JSKit. But the installation script section is helpful for new contributors.

MaxDesiatov:
That's updated, `README.md` now lists steps for newcomers that allow them to use `carton`, but keeps a separate section for advanced use cases and local development.

carson-katri (approved):

scripts/install-toolchain.sh
```diff
@@ -18,13 +18,16 @@ fi
 
 case $(uname -s) in
   Darwin)
-    toolchain_download="$swift_tag-osx.tar.gz"
+    toolchain_download="$swift_tag-macos-x86_64.tar.gz"
+    echo '-macos' >> .swift-version
```
kateinoigakukun, line 22: Do we still need this line?

MaxDesiatov: Still waiting for the new 5.3 toolchain to be tagged tonight that fixes this.

MaxDesiatov: That's cleaned up now, ready for review.

README.md
```diff
@@ -68,7 +50,82 @@ struct Pet: Codable {
...
+```sh
+$ swiftenv install https://github.com/swiftwasm/swift/releases/download/swift-wasm-5.3-SNAPSHOT-2020-10-16-a/swift-wasm-5.3-SNAPSHOT-2020-10-16-a-macos-x86_64.tar.gz
+```
+
+You can also use the `install-toolchain.sh` helper script that uses a hardcoded toolchain snapshot:
+
+```sh
+$ ./scripts/install-toolchain.sh
+$ swift --version
+Swift version 5.3-dev (LLVM 09686f232a, Swift 5a196c7f13)
```
kateinoigakukun, line 129: I think vendor name can be updated 👍

kateinoigakukun: I found that https://github.com/swiftwasm/swift/pull/1921 is not included in 5.3 branch 😢

kateinoigakukun (approved):

PR #102, opened 2020-10-22
JS Exception Support

kateinoigakukun:
Currently, there is no way to catch exceptions thrown by JavaScript functions using JavaScriptKit. 

This PR introduces new two APIs to catch exceptions. 

1. `JSFunction.throws`
2. `JSObject.throwing`

`JSFunction.throws` converts `JSFunction` into throwable function. `JSObject.throwing` provides throwing method call API capturing `this` context.

These two APIs have different names to avoid overload ambiguity. 

e.g.

```swift
// Throwing function call
let fn1: JSFunction = ...
try fn1.throws()

// Throwing constructor call
let Animal: JSFunction = ...
let cat1 = Animal.throws.new()

// Throwing method call
let obj1: JSObject = ...
obj1.throwing.method1!()
```

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -5,7 +5,7 @@ interface ExportedMemory {
 type ref = number;
 type pointer = number;
 
-interface GlobalVariable { }
+interface GlobalVariable {}
 declare const window: GlobalVariable;
 declare const global: GlobalVariable;
 let globalVariable: any;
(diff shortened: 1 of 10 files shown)


MaxDesiatov (approved):
LGTM, awesome work!

Runtime/src/index.ts
```diff
@@ -5,7 +5,7 @@ interface ExportedMemory {
 type ref = number;
 type pointer = number;
 
-interface GlobalVariable {}
+interface GlobalVariable { }
```
j-f1, line 8: Can you run Prettier? I’m pretty sure it removes this space

Runtime/src/index.ts
```diff
@@ -383,12 +385,18 @@ export class SwiftRuntime {
...
+                    result = Reflect.apply(
+                        func,
+                        undefined,
+                        decodeValues(argv, argc)
+                    );
+                } catch (error) {
+                    writeValue(error, kind_ptr, payload1_ptr, payload2_ptr, true);
+                    return;
+                }
+                writeValue(result, kind_ptr, payload1_ptr, payload2_ptr, false);
```
j-f1, line 399: Why can’t this be moved inside the `try`?

```js
writeValue(
  Reflect.apply(
    func,
    undefined,
    decodeValues(argv, argc)
  ),
  kind_ptr,
  payload1_ptr,
  payload2_ptr,
  false
);
```

Runtime/src/index.ts
```diff
@@ -420,6 +434,29 @@ export class SwiftRuntime {
...
+                result_obj: pointer,
+                exception_kind_ptr: pointer,
+                exception_payload1_ptr: pointer,
+                exception_payload2_ptr: pointer
+            ) => {
+                const obj = this.heap.referenceHeap(ref);
+                let result: any
+                try {
+                    result = Reflect.construct(obj, decodeValues(argv, argc));
+                    if (typeof result != "object")
```
j-f1, line 450: I’m pretty sure `new` can’t return a non-object. If you try to return a non-object in a constructor, it will use the `this` value instead.

See the spec [here](https://read262.jedfox.com/ordinary-and-exotic-objects-behaviours/ecmascript-function-objects#sec-ecmascript-function-objects-construct-argumentslist-newtarget), where step 12 is the most relevant.

Runtime/src/index.ts
kateinoigakukun: It's intentional. `writeValue` can throw exceptions but it's system errors and users don't want to catch those kinds of system errors.

kateinoigakukun:
What do you think about this API? I think there may be better interface.

MaxDesiatov:
If there's a better API for it, I can't come up with it...

j-f1 (approved):
LGTM! We’re not at 1.0 yet so we can continue to iterate if necessary.

PR #103, opened 2020-10-26
Check source code compatibility

kateinoigakukun:


.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,23 @@
+name: Check compatibility
+on:
+  pull_request:
+  push:
+    branches: [master]
```
MaxDesiatov, line 5: Should refer to `main`

MaxDesiatov (approved):
👍

PR #105, opened 2020-10-26
Fix branch and triple in `compatibility.yml`

MaxDesiatov:
`master` doesn't exist in our repository, so the workflow wasn't triggered. Also, the triple option had a typo, not sure how that even worked 🤔


kateinoigakukun (approved):
Sorry, thanks 😭

PR #112, opened 2021-01-02
Experimental async/await support

kateinoigakukun:
```swift
import JavaScriptEventLoop
import JavaScriptKit

JavaScriptEventLoop.install()
let fetch = JSObject.global.fetch.function!.async

func printZen() async {
  let result = await try! fetch("https://api.github.com/zen").object!
  let text = await try! result.asyncing.text!()
  print(text)
}

JavaScriptEventLoop.runAsync {
  await printZen()
}

```

## Current Issues

- [x] Many C++ headers are copied from LLVM and Swift repository even though the runtime uses only a few interfaces.
- [x] No tests
- [ ] Use of unsafe flag `-enable-experimental-concurrency`. If we remove it from Package.swift, users have to pass it from command line.


Sources/JavaScriptEventLoop/JSAsyncFunction.swift
```diff
@@ -0,0 +1,68 @@
...
+    private let base: JSObject
+    public init(_ base: JSObject) {
+        self.base = base
+    }
+
+    /// Returns the `name` member method binding this object as `this` context.
+    /// - Parameter name: The name of this object's member to access.
+    /// - Returns: The `name` member method binding this object as `this` context.
+    @_disfavoredOverload
+    public subscript(_ name: String) -> ((ConvertibleToJSValue...) async throws -> JSValue)? {
```
j-f1, line 48: Would it be possible to implement this as an overload in the regular `JSObject`? Not sure how overloads work exactly in this case but it would lead to a much nicer DX if possible.

kateinoigakukun: Unfortunately, `async` and non-`async` function overloads are ambiguous 😢

MaxDesiatov:
Do I understand correctly that after this is merged we'll require our users to use dev toolchains and to compile with experimental concurrency enabled?

MaxDesiatov:
Also, could we link with Swift runtime parts that are already included in the toolchain/SDK? Is a copy of it and LLVM stuff in our codebase required for this to work?

kateinoigakukun:
> Do I understand correctly that after this is merged we'll require our users to use dev toolchains and to compile with experimental concurrency enabled?

Yes, so I'm planning to merge this to `next` branch which will be used to use non-stable language features.

 > Also, could we link with Swift runtime parts that are already included in the toolchain/SDK? Is a copy of it and LLVM stuff in our codebase required for this to work?

`JavaScriptEventLoop` should be linked with `libswiftCore.a` and `libswift_Concurrency.a` to work. And LLVM headers are required to include `swift/ABI/Task.h`.

MaxDesiatov:
Fair enough. I'm only worried that contents of these headers can change in a breaking way, especially given that concurrency stuff isn't ABI-stable yet, and maybe there's a better way to track them. A submodule is probably not practical due to the size of repositories that have these headers, so I don't have a better alternative to propose... Maybe some kind of scripts that pull latest versions of these headers automatically? 🤔

kateinoigakukun:
Yes, so I wrote a shell script to fetch headers from `swiftwasm/swift` repository based on `.swift-version` snapshot.
https://github.com/swiftwasm/JavaScriptKit/pull/112/files#diff-306f08a700283b50101c5598d1724033f49fc492dbbdabb2c546067dbdc5bbe5R1-R74

MaxDesiatov (approved):
Seems legit then 👍

PR #113, opened 2021-01-02
Major API change on JSClosure

kateinoigakukun:
## Motivation

While integrating async/await feature, I found `JSPromise` has too complex API. (e.g. too many overloads, unsafe generic parameters, and lifetime management)
So I want to refine the API before async/await integration. As a first step, I changed `JSClosure` API.

## Changes

- Add `JSOneshotClosure` which provides an ability to invoke a closure at most once. This will be suitable to use with functions like `Promise.then`
- Remove `JSClosure.init(_ body: @escaping ([JSValue]) -> Void)` overload, which forces users to write result type in closure.
- Add JSClosure lifetime test suites


Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -178,123 +115,3 @@ fileprivate func invokeJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleT
...
-        // 3. Create a new JavaScript function which calls the given Swift function.
-        var objectRef: JavaScriptObjectRef = 0
-        _create_function(funcRef, &objectRef)
-
-        hostFuncRef = funcRef
-        id = objectRef
-    }
-    
-    /// A convenience initializer which assumes that the given body function returns `JSValue.undefined`
-    convenience public init(_ body: @escaping ([JSValue]) -> ()) {
```
MaxDesiatov, line 226: Can we keep this overload for at least one minor version as deprecated? This would give people some time to migrate away from it.

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -0,0 +1,127 @@
+import _CJavaScriptKit
+
+fileprivate var sharedFunctions: [JavaScriptHostFuncRef: ([JSValue]) -> JSValue] = [:]
+
+/// `JSOneshotClosure` is a JavaScript function that can be called at once.
```
MaxDesiatov, line 5:
```suggestion
/// `JSOneshotClosure` is a JavaScript function that can be called only once.
```

IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
```diff
@@ -183,6 +183,45 @@ try test("Function Call") {
...
+        _ = try expectThrow(try c1.throws())
+        c1.release()
+    }
+
+    do {
+        let c1 = JSOneshotClosure { _ in
+            return .boolean(true)
+        }
+        try expectEqual(c1(), .boolean(true))
+        // second call will cause fatalError that can be catched as a JavaScript exception
```
MaxDesiatov, line 219:
```suggestion
        // second call will cause `fatalError` that can be caught as a JavaScript exception
```

MaxDesiatov (commented):
Overall this seems legit with a few suggestions in comments.

MaxDesiatov (approved):
I wasn't sure about that `JSClosure.init` overload myself when I introduced, but you're right it brings more harm than good. Also, great idea to simplify a lot of things with `JSOneshotClosure`! 👍

kateinoigakukun:
@swiftwasm/jskit-team Could anyone review this change? I want two or more approvals for this kind of API change before merging  🙏

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -0,0 +1,140 @@
...
+// │          ┌ ─ ─ ─ ─ ─│─ ─│─ ─ ─ ─ ─ ┐         │ │
+// │          ↓          │   │          │         │ │
+// │  [Swift Closure]    │   │  Host Function ID  │ │
+// │          │          │   │                    │ │
+// │          │          │   └────────────────────┘ │
+// │          │          │                    ↑     │
+// │          │        Apply                  │     │
+// │          └─[Result]─┼───>[Callback func]─┘     │
+// │                     │                          │
+// └─────────────────────┴──────────────────────────┘
```
j-f1, line 123: I’ve usually seen this referred to as just a “thunk:”

```suggestion
// 1. Create a thunk in the JavaScript world, which has a reference
//    to a Swift closure.
// ┌─────────────────────┬──────────────────────────┐
// │     Swift side      │   JavaScript side        │
// │                     │                          │
// │                     │                          │
// │                     │   ┌──────[Thunk]───────┐ │
// │          ┌ ─ ─ ─ ─ ─│─ ─│─ ─ ─ ─ ─ ┐         │ │
// │          ↓          │   │          │         │ │
// │  [Swift Closure]    │   │  Host Function ID  │ │
// │                     │   │                    │ │
// │                     │   └────────────────────┘ │
// └─────────────────────┴──────────────────────────┘
//
// 2. When the thunk function is invoked, it calls the Swift closure via
//    `_call_host_function` and receives the result through a callback.
// ┌─────────────────────┬──────────────────────────┐
// │     Swift side      │   JavaScript side        │
// │                     │                          │
// │                     │                          │
// │                   Apply ┌──────[Thunk]───────┐ │
// │          ┌ ─ ─ ─ ─ ─│─ ─│─ ─ ─ ─ ─ ┐         │ │
// │          ↓          │   │          │         │ │
// │  [Swift Closure]    │   │  Host Function ID  │ │
// │          │          │   │                    │ │
// │          │          │   └────────────────────┘ │
// │          │          │                    ↑     │
// │          │        Apply                  │     │
// │          └─[Result]─┼───>[Callback func]─┘     │
// │                     │                          │
// └─────────────────────┴──────────────────────────┘
```

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -0,0 +1,140 @@
...
+    var isReleased: Bool = false
+
+    @available(*, deprecated, message: "This initializer will be removed in the next minor version update. Please use `init(_ body: @escaping ([JSValue]) -> JSValue)`")
+    @_disfavoredOverload
+    public init(_ body: @escaping ([JSValue]) -> ()) {
+        super.init({
+            body($0)
+            return .undefined
+        })
+    }
```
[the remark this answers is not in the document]
MaxDesiatov: The issue is with Swift typechecker which can't treat the currently available set of `JSClosure.init` overloads unambiguously without adding an explicit function signature at the place of use ([example one](https://github.com/TokamakUI/Tokamak/blob/a631d181e69e2e72d898b0b778356e61fb03e93b/Sources/TokamakDOM/App/ScenePhaseObserver.swift#L24),  [example two](https://github.com/TokamakUI/Tokamak/blob/3451d9ea12d28f060aa9fa84547e4fdacfc63a8d/Sources/TokamakTestRenderer/TestRenderer.swift#L27)). That defeats the purpose of the overload. I think we'd rather require users to add `return .undefined` explicitly (which can be easily copy-pasted where needed) instead of adding function signatures explicitly (which are almost always ad hoc).

Sources/JavaScriptKit/BasicObjects/JSTimer.swift
```diff
@@ -34,7 +34,10 @@ public final class JSTimer {
     - callback: the closure to be executed after a given `millisecondsDelay` interval.
   */
   public init(millisecondsDelay: Double, isRepeating: Bool = false, callback: @escaping () -> ()) {
-    closure = JSClosure { _ in callback() }
+    closure = JSClosure { _ in
```
j-f1, line 37: If `isRepeating` is false, this can be a one-shot closure, right?

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -52,7 +52,10 @@ public final class JSPromise<Success, Failure>: ConvertibleToJSValue, Constructi
     /** Schedules the `success` closure to be invoked on sucessful completion of `self`.
     */
     public func then(success: @escaping () -> ()) {
-        let closure = JSClosure { _ in success() }
+        let closure = JSClosure { _ in
```
j-f1, line 55: All promise closures can be one-shot, so I’m pretty sure we don’t need to hold on to them manually on the Swift side.

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -0,0 +1,140 @@
...
+        _create_function(funcRef, &objectRef)
+
+        hostFuncRef = funcRef
+        id = objectRef
+    }
+
+    public override func callAsFunction(this: JSObject? = nil, arguments: [ConvertibleToJSValue]) -> JSValue {
+        defer { release() }
+        return super.callAsFunction(this: this, arguments: arguments)
+    }
```
j-f1, line 27: I’m wondering — if the function will be destroyed as soon as it’s called, does it even make sense to call a one-shot closure from Swift?

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -0,0 +1,140 @@
...
+/// ...
+/// button.removeEventListener!("click", JSValue.function(eventListenter))
+/// eventListenter.release()
+/// ```
+///
+public class JSClosure: JSOneshotClosure {
+    
+    var isReleased: Bool = false
+
+    @available(*, deprecated, message: "This initializer will be removed in the next minor version update. Please use `init(_ body: @escaping ([JSValue]) -> JSValue)`")
```
j-f1, line 59:
```suggestion
    @available(*, deprecated, message: "This initializer will be removed in the next minor version update. Please use `init(_ body: @escaping ([JSValue]) -> JSValue)` and add `return .undefined` to the end of your closure")
```

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -0,0 +1,140 @@
...
+        isReleased = true
+        super.release()
+    }
+
+    deinit {
+        guard isReleased else {
+            fatalError("""
+            release() must be called on closures manually before deallocating.
+            This is caused by the lack of support for the `FinalizationRegistry` API in Safari.
+            """)
```
j-f1, line 86: I think this makes more sense as a comment
```suggestion
            // Safari doesn't support `FinalizationRegistry`, so we cannot automatically manage the lifetime of Swift objects
            fatalError("release() must be called on JSClosure objects manually before they are deallocated")
```

j-f1 (commented):
at a desktop now. LGTM except for these things

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
kateinoigakukun: Yes, usually users don't call `JSOneshotClosure` and `JSClosure` directly from Swift.
It might make sense to remove `JSFunction` inheritance and remove `callAsFunction`.

kateinoigakukun:
@j-f1 I addressed reviewed points. Could you take a look again?

Sources/JavaScriptKit/JSValue.swift
```diff
@@ -142,7 +142,12 @@ extension JSValue {
     /// ```
     @available(*, deprecated, message: "Please create JSClosure directly and manage its lifetime manually.")
     public static func function(_ body: @escaping ([JSValue]) -> JSValue) -> JSValue {
-        .function(JSClosure(body))
+        .object(JSClosure(body))
+    }
+
+    @available(*, deprecated, renamed: "object", message: "JSClosure is not a subclass of JSFunction now")
```
j-f1, line 148:
```suggestion
    @available(*, deprecated, renamed: "object", message: "JSClosure is no longer a subclass of JSFunction. Use .object(closure) instead.")
```

IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
```diff
@@ -488,21 +490,29 @@ try test("Date") {
...
     timeout = JSTimer(millisecondsDelay: timeoutMilliseconds, isRepeating: false) {
         // verify that at least `timeoutMilliseconds` passed since the `timeout` timer started
         try! expectEqual(start + timeoutMilliseconds <= JSDate().valueOf(), true)
     }
+    timeouts += [timeout]
+
+    timeout = JSTimer(millisecondsDelay: timeoutMilliseconds, isRepeating: false) {
+        fatalError("timer should be cancelled")
+    }
+    timeout = nil
```
j-f1, line 509: Totally happy to punt this to another issue, but since there isn’t necessarily a need to cancel  a `setTimeout`, could there be an API that doesn’t require holding the `JSTimer` instance? I guess users could just do 

```swift
JSObject.global.setTimeout(JSOneshotClosure {
  // ...
  return undefined
}, 1000)
```

kateinoigakukun: Yes, we can provide such API in this signature for example.

```swift
JSTimer.timeout(millisecondsDelay: 100) {
    // ...
}
```

This PR is already too large, so let's implement it in another PR 👍

jed-fox-neeva (approved):

j-f1 (approved):
lol oops accounts are hard

PR #115, opened 2021-01-04
Simplify JSPromise API

kateinoigakukun:
### Remove generic type parameters from `JSPromise`

These type parameters are type-safe only when it's produced by `JSPromise.then` and in other cases, it's always not type-safe. So I removed those type parameters.

### Remove `then` and `catch` overloads

Those overloads made it difficult to solve compile errors 😢

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
@@ -66,20 +66,20 @@ public final class JSPromise: JSBridgedClass {
         self.init(unsafelyWrapping: Self.constructor.new(closure))
     }
 
-    public static func resolve(_ value: JSValue) -> JSPromise {
+    public static func resolve(_ value: ConvertibleToJSValue) -> JSPromise {
         self.init(unsafelyWrapping: Self.constructor.resolve!(value).object!)
     }
 
-    public static func reject(_ reason: JSValue) -> JSPromise {
+    public static func reject(_ reason: ConvertibleToJSValue) -> JSPromise {
         self.init(unsafelyWrapping: Self.constructor.reject!(reason).object!)
     }
 
     /** Schedules the `success` closure to be invoked on sucessful completion of `self`.
     */
     @discardableResult
-    public func then(success: @escaping (JSValue) -> JSValue) -> JSPromise {
+    public func then(success: @escaping (JSValue) -> ConvertibleToJSValue) -> JSPromise {
         let closure = JSOneshotClosure {
-            return success($0[0])
+            return success($0[0]).jsValue()
         }
         return JSPromise(unsafelyWrapping: jsObject.then!(closure).object!)
     }
@@ -87,13 +87,13 @@ public final class JSPromise: JSBridgedClass {
     /** Schedules the `success` closure to be invoked on sucessful completion of `self`.
     */
     @discardableResult
-    public func then(success: @escaping (JSValue) -> JSValue,
-                     failure: @escaping (JSValue) -> JSValue) -> JSPromise {
+    public func then(success: @escaping (JSValue) -> ConvertibleToJSValue,
+                     failure: @escaping (JSValue) -> ConvertibleToJSValue) -> JSPromise {
         let successClosure = JSOneshotClosure {
-            return success($0[0])
+            return success($0[0]).jsValue()
         }
         let failureClosure = JSOneshotClosure {
-            return failure($0[0])
+            return failure($0[0]).jsValue()
         }
         return JSPromise(unsafelyWrapping: jsObject.then!(successClosure, failureClosure).object!)
     }
@@ -101,9 +101,9 @@ public final class JSPromise: JSBridgedClass {
     /** Schedules the `failure` closure to be invoked on rejected completion of `self`.
     */
     @discardableResult
-    public func `catch`(failure: @escaping (JSValue) -> JSValue) -> JSPromise {
+    public func `catch`(failure: @escaping (JSValue) -> ConvertibleToJSValue) -> JSPromise {
         let closure = JSOneshotClosure {
-            return failure($0[0])
+            return failure($0[0]).jsValue()
         }
         return .init(unsafelyWrapping: jsObject.catch!(closure).object!)
     }
--- a/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
@@ -46,7 +46,7 @@ public final class JSPromise: JSBridgedClass {
     /** Creates a new `JSPromise` instance from a given `resolver` closure. `resolver` takes
     two closure that your code should call to either resolve or reject this `JSPromise` instance.
     */
-    public convenience init(resolver: @escaping (@escaping (Result< JSValueConvertible, JSValueConvertible >) -> ()) -> ()) {
+    public convenience init(resolver: @escaping (@escaping (Result<JSValue, JSValue>) -> ()) -> ()) {
         let closure = JSOneshotClosure { arguments in
             // The arguments are always coming from the `Promise` constructor, so we should be
             // safe to assume their type here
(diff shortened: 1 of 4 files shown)


MaxDesiatov (commented):
While it's true that arbitrary user code returning a promise may not be type-safe, we know that most if not all promise-returning Web APIs can only fail with `JSError` and only resolve to a single type. If a user doesn't know what type to expect they can specify their expected type as `JSPromise<JSValue, JSValue>`.

While it could make sense to shift the reponsibility for dynamic casts to users, I'm not sure if it's an improvement on balance from the ergonomics perspective. This change would break a lot of things in OpenCombineJS and make `Publisher` conformance on promise-returning APIs harder to implement. For example, how would DOMKit promise-returning APIs like `fetch` look in DOMKit if this change is merged? I'm also interested to look what impact this has on `async`/`await` APIs in the end?

I don't have a strong opinion here to reject the change outright though. I'm interested to hear opinions on this from other people in the meantime.

kateinoigakukun:
@MaxDesiatov 
My explanation was too short, sorry.

My opinion is that JSPromise should be as primitive as possible. The current JSPromise has two responsibilities as a bridge to the Promise API and as a continuation monad.

Promise API: The responsibility for bridging Promise API is to provide an ability to register asynchronous task in the JavaScript event loop.

Continuation monad: The responsibility for continuation monad is to provide map and flatMap.

There are some problems when these two responsibilities are implemented on top of the Promise API.


1. For example, users usually want to use Promise as a continuation monad, but every time they chain in `then` (`map`), they need to interact with JavaScript world.
2. And also, the constraint of single type resolution is not something that Promise has, but is a convention of the individual JavaScript APIs. Therefore, JSPromise is more strongly constrained than JavaScript's Promise.

The main purpose of my change is to make JSPromise's responsibility only to bridge the Promise API, and to delegate the responsibility as a type-safe continuation monad to another type like Combine.Future.


In this way, the user doesn't need to interact with JavaScript to `map`. This separation of the two responsibilities results removing generic type parameters.


>  For example, how would DOMKit promise-returning APIs like fetch look in DOMKit if this change is merged? 

Libraries that provide type-safe APIs, such as DOMKit, should not return JSPromise directly to the user, but convert it to a type that provides type-safe continuation monad like Combine.Future.

> I'm also interested to look what impact this has on async/await APIs in the end?

This doesn't directly affect the async/await APIs, it's just a point I noticed while implementing the Concurrency feature.

MaxDesiatov (approved):
Thanks for the explanation. This makes sense 👍

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -40,73 +43,10 @@ public final class JSPromise<Success, Failure>: ConvertibleToJSValue, Constructi
...
-        self.init(unsafe: JSObject.global.Promise.function!.new(closure))
-    }
-}
-
-extension JSPromise where Success: ConvertibleToJSValue, Failure: JSError {
-    /** Creates a new `JSPromise` instance from a given `resolver` closure. `resolver` takes 
-    a closure that your code should call to either resolve or reject this `JSPromise` instance.
-    */
-    public convenience init(resolver: @escaping (@escaping (Result<Success, JSError>) -> ()) -> ()) {
+    public convenience init(resolver: @escaping (@escaping (Result<JSValue, JSValue>) -> ()) -> ()) {
```
j-f1, line 49: Could these be `JSValueConvertible` so users can do something like `resolve(.success("Hello, world!"))`?
```suggestion
    public convenience init(resolver: @escaping (@escaping (Result< JSValueConvertible, JSValueConvertible >) -> ()) -> ()) {
```

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -116,123 +56,67 @@ extension JSPromise where Success: ConvertibleToJSValue, Failure: JSError {
...
-    ) {
-        let closure = JSOneshotClosure { arguments in
-            guard let result = Success.construct(from: arguments[0]) else {
-                fatalError("\(file):\(line): failed to unwrap success value for `then` callback")
-            }
-            success(result)
-            return .undefined
-        }
-        _ = jsObject.then!(closure)
+    public static func resolve(_ value: JSValue) -> JSPromise {
```
j-f1, line 69:
```suggestion
    public static func resolve(_ value: JSValueConvertible) -> JSPromise {
```

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -116,123 +56,67 @@ extension JSPromise where Success: ConvertibleToJSValue, Failure: JSError {
...
-        line: Int = #line
-    ) -> JSPromise<ResultType, Failure> {
-        let closure = JSOneshotClosure { arguments -> JSValue in
-            guard let result = Success.construct(from: arguments[0]) else {
-                fatalError("\(file):\(line): failed to unwrap success value for `then` callback")
-            }
-            return success(result).jsValue()
-        }
-        return .init(unsafe: jsObject.then!(closure).object!)
+    public static func reject(_ reason: JSValue) -> JSPromise {
```
j-f1, line 73:
```suggestion
    public static func reject(_ reason: JSValueConvertible) -> JSPromise {
```

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -116,123 +56,67 @@ extension JSPromise where Success: ConvertibleToJSValue, Failure: JSError {
...
-        file: StaticString = #file,
-        line: Int = #line
-    ) -> JSPromise<ResultSuccess, ResultFailure> {
-        let closure = JSOneshotClosure { arguments -> JSValue in
-            guard let result = Success.construct(from: arguments[0]) else {
-                fatalError("\(file):\(line): failed to unwrap success value for `then` callback")
-            }
-            return success(result).jsValue()
+    @discardableResult
+    public func then(success: @escaping (JSValue) -> JSValue) -> JSPromise {
```
j-f1, line 80:
```suggestion
    public func then(success: @escaping (JSValue) -> JSValueConvertible) -> JSPromise {
```

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -116,123 +56,67 @@ extension JSPromise where Success: ConvertibleToJSValue, Failure: JSError {
...
-        line: Int = #line
-    ) -> JSPromise<ResultSuccess, Never> {
-        let closure = JSOneshotClosure { arguments -> JSValue in
-            guard let error = Failure.construct(from: arguments[0]) else {
-                fatalError("\(file):\(line): failed to unwrap error value for `catch` callback")
-            }
-            return failure(error).jsValue()
+    @discardableResult
+    public func then(success: @escaping (JSValue) -> JSValue,
+                     failure: @escaping (JSValue) -> JSValue) -> JSPromise {
```
j-f1, line 91:
```suggestion
    public func then(success: @escaping (JSValue) -> JSValueConvertible,
                     failure: @escaping (JSValue) -> JSValueConvertible) -> JSPromise {
```

I wonder if there’s a way to implement this as `Result<JSValue, JSValue>`? Maybe a disfavored overload so the user has to explicitly specify the label?

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
```diff
@@ -116,123 +56,67 @@ extension JSPromise where Success: ConvertibleToJSValue, Failure: JSError {
...
-        line: Int = #line
-    ) {
-        let closure = JSOneshotClosure { arguments in
-            guard let error = Failure.construct(from: arguments[0]) else {
-                fatalError("\(file):\(line): failed to unwrap error value for `catch` callback")
-            }
-            failure(error)
-            return .undefined
+    @discardableResult
+    public func `catch`(failure: @escaping (JSValue) -> JSValue) -> JSPromise {
```
j-f1, line 104:
```suggestion
    public func `catch`(failure: @escaping (JSValue) -> JSValueConvertible) -> JSPromise {
```

Sources/JavaScriptKit/JSBridgedType.swift
```diff
@@ -35,7 +35,12 @@ public protocol JSBridgedClass: JSBridgedType {
...
     public init?(from value: JSValue) {
-        guard let object = value.object, object.isInstanceOf(Self.constructor) else { return nil }
+        guard let object = value.object else { return nil }
+        self.init(from: object)
+    }
+
+    public init?(from object: JSObject) {
+        guard object.isInstanceOf(Self.constructor) else { return nil }
         self.init(unsafelyWrapping: object)
     }
```
j-f1, line 45: Nice!

Sources/JavaScriptKit/BasicObjects/JSPromise.swift
kateinoigakukun: `ConvertibleToJSValue` is not conforming to `Error`, so `Result` can't have Failure as `ConvertibleToJSValue` :cry:

kateinoigakukun:
@j-f1 Could you take a look again?

j-f1 (approved):
Sorry for the delay!

PR #128, opened 2021-04-29
Refactor JSClosure to leverage `FinalizationRegistry`

### LINKED ISSUE #106: Avoid manual memory management with JSClosure
I wonder if [reference types](https://bytecodealliance.org/articles/reference-types-in-wasmtime) could lift our requirement to manage references to JS closures manually?

> With `externref`s, the host gives Wasm limited access to host objects, but the host also needs to know when Wasm is finished with those objects, so it can clean them up. That clean up might involve closing a file handle or simply deallocating memory. 

This means that JS closures should be passed to the Swift code as `externref`. Deallocating all references to `externref` would deallocate the closure on the JS side.

Reference types are now available in all browsers except Safari, and in the latter case they seem to only be hidden behind `JSC_useWebAssemblyReferences` flag. The only question is `externref` support in our toolchain...

Doesn't look like LLVM had much progress in that direction? 🤔 Also see https://github.com/WebAssembly/tool-conventions/issues/122

j-f1:
- `closure.release()` and `JSOneshotClosure` are no longer needed
- `FinalizationRegistry` calls `swjs_free_host_function` to remove the corresponding Swift closure from the global dictionary.
- deallocating a `JSClosure` removes the JS function from the heap. If the function isn’t being held somewhere else (such as inside an event handler), it will become eligible for garbage collection, and the Swift closure will be eventually released.
- `JavaScriptHostFuncRef` is now obtained from a global “expando” variable because `JSClosure { ... }; JSClosure { ... }` may reuse an `ObjectIdentifier`. If we decide to add multithreading, we should be careful to only modify `closureRef` using a lock.
- TypeScript has been updated to add support for `WeakRef`/`FinalizationRegistry`

Fixes #106.

### DIFF
--- a/Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
+++ b/Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
@@ -1,5 +1,6 @@
 import _CJavaScriptKit
 
+fileprivate var closureRef: JavaScriptHostFuncRef = 0
 fileprivate var sharedClosures: [JavaScriptHostFuncRef: ([JSValue]) -> JSValue] = [:]
 
 /// JSClosureProtocol abstracts closure object in JavaScript, whose lifetime is manualy managed
@@ -10,40 +11,9 @@ public protocol JSClosureProtocol: JSValueCompatible {
     func release()
 }
 
-/// `JSOneshotClosure` is a JavaScript function that can be called only once.
-public class JSOneshotClosure: JSObject, JSClosureProtocol {
-    private var hostFuncRef: JavaScriptHostFuncRef = 0
-
-    public init(_ body: @escaping ([JSValue]) -> JSValue) {
-        // 1. Fill `id` as zero at first to access `self` to get `ObjectIdentifier`.
-        super.init(id: 0)
-        let objectId = ObjectIdentifier(self)
-        let funcRef = JavaScriptHostFuncRef(bitPattern: Int32(objectId.hashValue))
-        // 2. Retain the given body in static storage by `funcRef`.
-        sharedClosures[funcRef] = {
-            defer { self.release() }
-            return body($0)
-        }
-        // 3. Create a new JavaScript function which calls the given Swift function.
-        var objectRef: JavaScriptObjectRef = 0
-        _create_function(funcRef, &objectRef)
-
-        hostFuncRef = funcRef
-        id = objectRef
-    }
-
-    /// Release this function resource.
-    /// After calling `release`, calling this function from JavaScript will fail.
-    public func release() {
-        sharedClosures[hostFuncRef] = nil
-    }
-}
 
 /// `JSClosure` represents a JavaScript function the body of which is written in Swift.
 /// This type can be passed as a callback handler to JavaScript functions.
-/// Note that the lifetime of `JSClosure` should be managed by users manually
-/// due to GC boundary between Swift and JavaScript.
-/// For further discussion, see also [swiftwasm/JavaScriptKit #33](https://github.com/swiftwasm/JavaScriptKit/pull/33)
 ///
 /// e.g.
 /// ```swift
@@ -55,12 +25,10 @@ public class JSOneshotClosure: JSObject, JSClosureProtocol {
 /// button.addEventListener!("click", JSValue.function(eventListenter))
 /// ...
 /// button.removeEventListener!("click", JSValue.function(eventListenter))
-/// eventListenter.release()
 /// ```
 ///
 public class JSClosure: JSObject, JSClosureProtocol {
     private var hostFuncRef: JavaScriptHostFuncRef = 0
-    var isReleased: Bool = false
 
     @available(*, deprecated, message: "This initializer will be removed in the next minor version update. Please use `init(_ body: @escaping ([JSValue]) -> JSValue)` and add `return .undefined` to the end of your closure")
     @_disfavoredOverload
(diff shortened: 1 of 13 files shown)


MaxDesiatov:
I wonder if there's a good way to feature-flag it? It would be great to dynamically detect if `FinalizationRegistry` is available. Otherwise merging this means projects depending on JSKit will break on browsers older than Safari 14.1.

j-f1:
I don’t think so — it would require substantially different API usage on the Swift side if it was unavailable unless we just wanted to leak memory on the Swift side. I’m not too concerned about compatibility because recent versions of most evergreen browsers support it, and Safari 14.1 is available on Mojave (10.14) and above, so most people should be OK with this, especially given the rate at which people are using SwiftWasm right now 😉

MaxDesiatov:
I understand, my concern is with economic accessibility here. There are enough old devices that will never get Safari 14.1 installed on them, and I would love Tokamak to support them. We also know some people researching Tokamak for use in production environment right now, that puts some constraints on supported browsers and devices too.

j-f1:
I guess I’m not sure what the target audience would be. Mixpanel’s [iOS adoption report](https://mixpanel.com/trends/#report/ios_14/from_date:-87,report_unit:week,to_date:0) shows that 90% of devices run iOS 14, and only 5% of devices run a version below iOS 13. Chrome seems to run on Android 5 and above (although I’m not 100% positive they will actually give you the latest version).

Two other things to consider: First, SwiftWasm’s current build output is fairly large, making it difficult to use on mobile data plans or with cheaper phones. Second, I doubt the pool of devices that support SwiftWasm but not FinalizationRegistry is growing.

MaxDesiatov:
After asking around, I was surprised to hear that requiring latest iOS is not such a problem for some devs. I'm happy to defer to Yuta on this after all.

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -10,40 +11,9 @@ public protocol JSClosureProtocol: JSValueCompatible {
     func release()
 }
 
-/// `JSOneshotClosure` is a JavaScript function that can be called only once.
-public class JSOneshotClosure: JSObject, JSClosureProtocol {
```
kateinoigakukun, line 14: Could you measure runtime performance between `JSOneshotClosure` and `JSClosure` for deallocation overhead?
I think deallocation that depends on GC cycle will increase memory usage while waiting until event loop frame. 

(But I'm not sure how this overhead affects users experience 😅 )

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -72,33 +40,29 @@ public class JSClosure: JSObject, JSClosureProtocol {
...
-        id = objectRef
+        super.init(id: objectRef)
     }
 
-    public func release() {
-        isReleased = true
-        sharedClosures[hostFuncRef] = nil
-    }
+    @available(*, deprecated, message: "JSClosure.release() is no longer necessary")
+    public func release() {}
```
kateinoigakukun, line 57: So `JSClosure` in browsers without `FinalizationRegistry` support won't be deallocated, right?

kateinoigakukun (commented):
I want to merge change with feature flag like `-D JAVASCRIPTKIT_WITHOUT_WEAKREF` as an escape hatch.  What do you think about this?

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
j-f1: If `FinalizationRegistry` is not supported, the JS side will crash on startup when it attempts to call `new FinalizationRegistry()`.

j-f1:
I‘d be happy to merge with a feature flag, but I’m not sure how to implement it on the JS side. I guess I could conditionally construct the `FinalizationRegistry` if a special API call (like the existing `version` call) returns `true`

kateinoigakukun:
@j-f1 I think we can expose an API to get feature bit-flags from Swift side like: `func swjs_feature_flags() -> UInt32`

nmn:
Just chiming in:

Instead of a flag, there should a separate API for `NoGCJSClosure`. If someone really wants to support older browsers, they'll need to write different Swift code anyway, so let them choose a different API.

Further, I don't think this is really a concern at all. In a few short months Safari 15 will be out Safari 14.1 will be the "old" version. It should be considered very safe to assume minimum Safari 14.1

MaxDesiatov:
Think I'm fine in proceeding with this approach after all without supporting older browsers. @kateinoigakukun do you think switching to `FinalizationRegistry` would have any impact on `async`/`await` support? Or do we need to prove that `FinalizationRegistry` does not have any performance impact first?

j-f1:
My recent updates to the JS side cause the `FinalizationRegistry` to always be initialized — if the plan is to drop support for older browsers I think [`c77e069` (#128)](https://github.com/swiftwasm/JavaScriptKit/pull/128/commits/c77e069c2d9d8b819a9e2649f6a8643d07c0f43f) can be reverted to officially remove the legacy closures. Otherwise, I’d have to make some changes to ensure the code continues to run on older runtimes.

MaxDesiatov:
@j-f1 what do you think would be the best to benchmark how `FinalizationRegistry` behaves against older closures API? Do we already have any benchmarks for either of these?

j-f1:
I don’t think benchmarks are the right way to evaluate this change. The primary result is to improve safety/ease of use by tying the lifetime of the Swift closure to that of the JS function. Since the benchmarks run above show no major performance changes, I’m not personally super concerned.

MaxDesiatov:
Right, I was only interested if it was worth dropping manually-managed version altogether in favor of the garbage-collected one. @kateinoigakukun WDYT?

kateinoigakukun:
Could you try this branch on Tokamak and ensure no compatibility issues?

kateinoigakukun (approved):

Issue #129, opened 2021-05-06
`make test` crashes due to `JSClosure` memory issues


MaxDesiatov:
I'm not sure why this isn't reproducible on CI, but I got it crashing off the latest code in the `main` branch with a simple `make && make test`, here's the full output:

Here's the full output:

```
cd IntegrationTests && \
	    CONFIGURATION=debug make test && \
	    CONFIGURATION=release make test
cd .. && npm run build

> javascript-kit-swift@0.10.1 build
> npm run build:clean && npm run build:ts


> javascript-kit-swift@0.10.1 build:clean
> rm -rf Runtime/lib


> javascript-kit-swift@0.10.1 build:ts
> cd Runtime; tsc -b

swift build --package-path TestSuites \
	            --product PrimaryTests \
	            --triple wasm32-unknown-wasi \
	            --configuration debug
[1/1] Linking PrimaryTests.wasm
mkdir -p dist
cp TestSuites/.build/debug/PrimaryTests.wasm dist/PrimaryTests.wasm
node bin/primary-tests.js
Fatal error: The function was already released: file JavaScriptKit/JSClosure.swift, line 143

Fatal error: release() must be called on JSClosure objects manually before they are deallocated: file JavaScriptKit/JSClosure.swift, line 97

Fatal error: The function was already released: file JavaScriptKit/JSClosure.swift, line 143

rm TestSuites/.build/debug/PrimaryTests.wasm
cd .. && npm run build

> javascript-kit-swift@0.10.1 build
> npm run build:clean && npm run build:ts


> javascript-kit-swift@0.10.1 build:clean
> rm -rf Runtime/lib


> javascript-kit-swift@0.10.1 build:ts
> cd Runtime; tsc -b

swift build --package-path TestSuites \
	            --product PrimaryTests \
	            --triple wasm32-unknown-wasi \
	            --configuration release
[1/1] Linking PrimaryTests.wasm
mkdir -p dist
cp TestSuites/.build/release/PrimaryTests.wasm dist/PrimaryTests.wasm
node bin/primary-tests.js
Fatal error: The function was already released: file JavaScriptKit/JSClosure.swift, line 143

Fatal error: release() must be called on JSClosure objects manually before they are deallocated: file JavaScriptKit/JSClosure.swift, line 97

Fatal error: The function was already released: file JavaScriptKit/JSClosure.swift, line 143

rm TestSuites/.build/release/PrimaryTests.wasm
```

Issue #132, opened 2021-07-16
Cannot set HTMLElement.style properties


Feuermurmel:
I have this really simple problem I can't seem to figure out. I'm trying to use `JavaScriptKit` to create an element in the browsers DOM and set some CSS style on it:

```
import JavaScriptKit

let document = JSObject.global.document

var spanElem = document.createElement("span")
spanElem.innerText = "hello"
spanElem.style.backgroundColor = "yellow"

_ = document.body.appendChild(spanElem)
```

But it seems that the assignment to `spanElem.style.backgroundColor` doesn't have any effect. When I run this code (as `main.swift`), the `<span>` is created but its CSS style is unchanged. Reading `style.backgroundColor` from the browser console also shows nothing, but changing it from there works as expected:

```
document.getElementsByTagName('span')[0].style.backgroundColor
""
document.getElementsByTagName('span')[0].style.backgroundColor = "yellow"
"yellow"
document.getElementsByTagName('span')[0].style.backgroundColor
"yellow"
```
I'm using carton 0.10.0, Swift swift-wasm-5.3.1-RELEASE, JavaScriptKit 0.10.1 and am testing with newest Chrome and Safari (with same results) on macOS 10.15.7.

What am I doing wrong here?

kateinoigakukun:
This is a limitation of the current API design of JavaScriptKit.
`spanElem.style.backgroundColor = "yellow"` is evaluated as below:
```swift
var style = spanElem.style
style.backgroundColor = "yellow"
spanElem.style = style
```
Even `style` is a reference type in JavaScript world, it's expressed by value type in Swift, so `spanElem.style = style` re-assignment is appeared. But this assignment causes undefined behavior because [`HTMLElement.style` is readonly property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style) and it clears all CSS attributes. 

I'll re-design JavaScriptKit API to resolve this semantics mismatch, but please use this workaround to avoid re-assignment for a while. Sorry.

```swift
import JavaScriptKit

let document = JSObject.global.document

var spanElem = document.createElement("span")
spanElem.innerText = "hello"
var style  = spanElem.style
style.backgroundColor = "yellow"

_ = document.body.appendChild(spanElem)
```

Feuermurmel:
Thank you for your very helpful response! With it I'm no longer stuck and can continue with my project. :) One note:

[...]

PR #133, opened 2021-07-18
Gardening tests

kateinoigakukun:
- Close https://github.com/swiftwasm/JavaScriptKit/issues/129 because it's an expected crash output
- Add a test case for https://github.com/swiftwasm/JavaScriptKit/issues/132

### DIFF
--- a/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
+++ b/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
@@ -186,6 +186,12 @@ try test("Function Call") {
 let evalClosure = JSObject.global.globalObject1.eval_closure.function!
 
 try test("Closure Lifetime") {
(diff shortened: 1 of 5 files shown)


IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
```diff
@@ -200,7 +200,9 @@ try test("Closure Lifetime") {
         }
         c1.release()
         // Call a released closure
+        print("======= BEGIN OF EXPECTED FATAL ERROR =====")
         _ = try expectThrow(try evalClosure.throws(c1))
```
MaxDesiatov, line 203: Should expectThrow contain these prints to avoid repetition?

kateinoigakukun: `expectThrow` is not for crash tests, so I added another helper function

MaxDesiatov (approved):
Seems legit, thanks!

PR #135, opened 2021-08-08
Test with multiple toolchain versions

kateinoigakukun:


MaxDesiatov (approved):

MaxDesiatov:
@kateinoigakukun I guess we should also test with Ubuntu 20.04, but I don't see where the exact Ubuntu version is passed from the matrix. I don't think `swiftenv` detects that automatically, does it?

kateinoigakukun:
@MaxDesiatov swiftenv doesn't anything around arch but install script does it https://github.com/swiftwasm/JavaScriptKit/blob/5d7c353e37456eed746cb142bdd28652d8a6acfd/scripts/install-toolchain.sh#L25-L29

PR #137, opened 2021-09-01
Support async/await and require Swift 5.5

MaxDesiatov:


kateinoigakukun:
Things have been changed at all, so I need to make it re-work again 😅

MaxDesiatov:
Fair enough, would it make sense to close it then?

MaxDesiatov:
I would love to work on this, I only hoped it could be done quickly 😅 But since it needs a rewrite, I will probably be too slow with this. Let's keep it open for now to show our intent if anyone wants to know about our plans for this. Can close this one when a separate branch/PR for a rewrite is needed.

MaxDesiatov:
@kateinoigakukun have you seen error messages in CI runs here? They were caused by `swiftasync` CC, which is apparently not supported by Clang when targeting Wasm. Is this no longer an issue in your replacement PR?

kateinoigakukun:
@MaxDesiatov 

At first, my new PR doesn't bring headers from toolchain, so there is no issue around it.

Then, think about the issue here.
I think that error will be gone after updating the Config.h to be the latest version.
it tries to use swiftasynccall in JSKit, but not supported on wasm https://github.com/swiftwasm/JavaScriptKit/pull/137/files#diff-e748bb05e91769e7ca05e9fd4ea4b4ff6b53c183d38d7b455fc4e52eb5860572R181-R185
On the toolchain side, the attribute is not used due to `__has_feature(swiftasynccc)` condition.
https://github.com/swiftwasm/swift/blob/b1d209396b535d4af0613ff1aa5deebd94e329a3/include/swift/Runtime/Config.h#L207

PR #139, opened 2021-09-09
Suggestion for #128: Add escape hatch for old environments

kateinoigakukun:
My only concern about #128 is dropping older versions of environments. The problem can be resolved easily and doesn't have much complexity.

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -125,12 +125,10 @@ class SwiftRuntimeHeap {
 /// Memory lifetime of closures in Swift are managed by Swift side
 class SwiftClosureHeap {
     private functionRegistry: FinalizationRegistry<number>;
-    private exports: SwiftRuntimeExportedFunctions
 
     constructor(exports: SwiftRuntimeExportedFunctions) {
-        this.exports = exports
         this.functionRegistry = new FinalizationRegistry((id) => {
-            this.exports.swjs_free_host_function(id);
+            exports.swjs_free_host_function(id);
         });
     }
 
@@ -160,24 +158,26 @@ export class SwiftRuntime {
         }
     }
     get closureHeap(): SwiftClosureHeap | null {
-        if (this._closureHeap)
-            return this._closureHeap;
+        if (this._closureHeap) return this._closureHeap;
         if (!this.instance)
             throw new Error("WebAssembly instance is not set yet");
 
         const exports = (this.instance
             .exports as any) as SwiftRuntimeExportedFunctions;
         const features = exports.swjs_library_features();
-        const librarySupportsWeakRef = (features & LibraryFeatures.WeakRefs) != 0;
+        const librarySupportsWeakRef =
+            (features & LibraryFeatures.WeakRefs) != 0;
         if (librarySupportsWeakRef) {
             if (typeof FinalizationRegistry !== "undefined") {
                 this._closureHeap = new SwiftClosureHeap(exports);
-		return this._closureHeap;
+                return this._closureHeap;
             } else {
-                throw new Error("The JavaScriptKit in Swift expects the target environment supports WeakRefs. Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable features using WeakRefs.");
+                throw new Error(
+                    "The Swift part of JavaScriptKit was configured to require the availability of JavaScript WeakRefs. Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable features that use WeakRefs."
+                );
             }
         }
-	return null;
+        return null;
     }
 
     importObjects() {
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -21,6 +21,7 @@ if (typeof globalThis !== "undefined") {
 
 interface SwiftRuntimeExportedFunctions {
     swjs_library_version(): number;
+    swjs_library_features(): number;
     swjs_prepare_host_function_call(size: number): pointer;
     swjs_cleanup_host_function_call(argv: pointer): void;
     swjs_call_host_function(
@@ -42,6 +43,10 @@ enum JavaScriptValueKind {
     Null = 4,
     Undefined = 5,
     Function = 6,
+}
+
+enum LibraryFeatures {
+    WeakRefs = 1 << 0,
 }
 
 type TypedArray =
@@ -117,25 +122,33 @@ class SwiftRuntimeHeap {
     }
 }
 
+/// Memory lifetime of closures in Swift are managed by Swift side
+class SwiftClosureHeap {
+    private functionRegistry: FinalizationRegistry<number>;
+    private exports: SwiftRuntimeExportedFunctions
+
+    constructor(exports: SwiftRuntimeExportedFunctions) {
+        this.exports = exports
+        this.functionRegistry = new FinalizationRegistry((id) => {
+            this.exports.swjs_free_host_function(id);
+        });
+    }
+
+    alloc(func: any, func_ref: number) {
+        this.functionRegistry.register(func, func_ref);
+    }
+}
+
 export class SwiftRuntime {
     private instance: WebAssembly.Instance | null;
     private heap: SwiftRuntimeHeap;
-    private functionRegistry: FinalizationRegistry<unknown>;
+    private _closureHeap: SwiftClosureHeap | null;
     private version: number = 701;
 
     constructor() {
         this.instance = null;
         this.heap = new SwiftRuntimeHeap();
-        this.functionRegistry = new FinalizationRegistry(
-            this.handleFree.bind(this)
-        );
-    }
-
-    handleFree(id: unknown) {
-        if (!this.instance || typeof id !== "number") return;
-        const exports = (this.instance
-            .exports as any) as SwiftRuntimeExportedFunctions;
-        exports.swjs_free_host_function(id);
+        this._closureHeap = null;
     }
 
     setInstance(instance: WebAssembly.Instance) {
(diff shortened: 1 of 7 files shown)


MaxDesiatov (approved):
Overall looks great, thanks!

kateinoigakukun:
Hmm, which line introduced performance regression... 🤔

MaxDesiatov:
Woah, yeah, that seems pretty serious

kateinoigakukun:
The serious regression appeared only on JavaScript cases. So no regression in our codebase, I hope.

MaxDesiatov:
What are you referring to by "JavaScript cases" here?

kateinoigakukun:
Here https://github.com/swiftwasm/JavaScriptKit/blob/43a33eed17c13ba575b30c722679a8134e7ef155/IntegrationTests/bin/benchmark-tests.js#L25-L40

MaxDesiatov:
So you mean that `FinalizationRegistry` turns out to be about 20% slower when using it in Node.js than relying on manual deallocation, and there's nothing we can do in our codebase to make that go faster?

kateinoigakukun:
I guess that's right. But not sure why this regression didn't happen on the `finalization-registry` branch 🤔

kateinoigakukun:
I couldn't see considerable regression on my end

MaxDesiatov:
Maybe it was a temporary CI glitch then?

Runtime/src/index.ts
```diff
@@ -117,25 +122,33 @@ class SwiftRuntimeHeap {
...
 
+/// Memory lifetime of closures in Swift are managed by Swift side
+class SwiftClosureHeap {
+    private functionRegistry: FinalizationRegistry<number>;
+    private exports: SwiftRuntimeExportedFunctions
+
+    constructor(exports: SwiftRuntimeExportedFunctions) {
+        this.exports = exports
+        this.functionRegistry = new FinalizationRegistry((id) => {
+            this.exports.swjs_free_host_function(id);
```
j-f1, line 133: Why not reference `exports` directly, removing the need for the `exports` property?
```suggestion
            exports.swjs_free_host_function(id);
```

Runtime/src/index.ts
```diff
@@ -117,25 +122,33 @@ class SwiftRuntimeHeap {
...
+    alloc(func: any, func_ref: number) {
+        this.functionRegistry.register(func, func_ref);
+    }
+}
+
 export class SwiftRuntime {
     private instance: WebAssembly.Instance | null;
     private heap: SwiftRuntimeHeap;
-    private functionRegistry: FinalizationRegistry<unknown>;
+    private _closureHeap: SwiftClosureHeap | null;
```
j-f1, line 145: Why the leading underscore when the other private properties don’t have it?

Runtime/src/index.ts
```diff
@@ -146,6 +159,26 @@ export class SwiftRuntime {
...
+            throw new Error("WebAssembly instance is not set yet");
+
+        const exports = (this.instance
+            .exports as any) as SwiftRuntimeExportedFunctions;
+        const features = exports.swjs_library_features();
+        const librarySupportsWeakRef = (features & LibraryFeatures.WeakRefs) != 0;
+        if (librarySupportsWeakRef) {
+            if (typeof FinalizationRegistry !== "undefined") {
+                this._closureHeap = new SwiftClosureHeap(exports);
+		return this._closureHeap;
```
j-f1, line 175: check your whitespace here — it looks like tabs got used instead of spaces

Runtime/src/index.ts
```diff
@@ -146,6 +159,26 @@ export class SwiftRuntime {
...
+            } else {
+                throw new Error("The JavaScriptKit in Swift expects the target environment supports WeakRefs. Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable features using WeakRefs.");
```
j-f1, line 177:
```suggestion
                throw new Error("The Swift part of JavaScriptKit was configured to require the availability of JavaScript WeakRefs. Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable features that use WeakRefs.");
```

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -1,9 +1,6 @@
 import _CJavaScriptKit
 
-fileprivate var closureRef: JavaScriptHostFuncRef = 0
-fileprivate var sharedClosures: [JavaScriptHostFuncRef: ([JSValue]) -> JSValue] = [:]
-
-/// JSClosureProtocol abstracts closure object in JavaScript, whose lifetime is manualy managed
+/// JSClosureProtocol abstracts closure object in JavaScript, whose lifetime might be manualy managed
```
j-f1, line 3: a little more thorough but perhaps not entirely correct:

```suggestion
/// JSClosureProtocol wraps Swift closure objects for use in JavaScript. Conforming types are responsible for managing the lifetime of the closure they wrap, but can delegate that task to the user by requiring an explicit `release()` call.
```

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -40,26 +73,27 @@ public class JSClosure: JSObject, JSClosureProtocol {
...
-        self.hostFuncRef = closureRef
-        closureRef += 1
-
-        // Retain the given body in static storage by `closureRef`.
-        sharedClosures[self.hostFuncRef] = body
-
-        // Create a new JavaScript function which calls the given Swift function.
+        // 1. Fill `id` as zero at first to access `self` to get `ObjectIdentifier`.
+        super.init(id: 0)
+        let objectId = ObjectIdentifier(self)
```
j-f1, line 78: I switched away from using `ObjectIdentifier` because:

```swift
foo(JSClosure { _ in "this closure" })
bar(JSClosure { _ in "and this closure" })
```

…ended up having the same `ObjectIdentifier` value. The monotonically increasing global ID works around that.

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -114,66 +148,36 @@ func _call_host_function_impl(
...
-            defer { self.release() }
-            return body($0)
-        }
-        // 3. Create a new JavaScript function which calls the given Swift function.
-        var objectRef: JavaScriptObjectRef = 0
-        _create_function(funcRef, &objectRef)
+// [WeakRefs](https://github.com/tc39/proposal-weakrefs) is already Stage 4, but it's still newish API,
+// so provide a escape hatch.
+// Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` by SwiftPM build system
+#if JAVASCRIPTKIT_WITHOUT_WEAKREFS
```
j-f1, line 156:
```suggestion
/// [WeakRefs](https://github.com/tc39/proposal-weakrefs) are already Stage 4,
/// but was added recently enough that older browser versions don’t support it.
/// Build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable the relevant behavior.
#if JAVASCRIPTKIT_WITHOUT_WEAKREFS
```

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -114,66 +148,36 @@ func _call_host_function_impl(
...
-        hostFuncRef = funcRef
-        id = objectRef
-    }
+extension JSClosure {
 
-    public func release() {
-        isReleased = true
-        sharedClosures[hostFuncRef] = nil
-    }
+    @available(*, deprecated, message: "JSClosure.release() is no longer necessary if the target environment supports WeakRefs")
```
j-f1, line 174: IMO this deprecation message implies some sort of feature checking has happened to determine that WeakRefs are supported.

```suggestion
    @available(*, deprecated, message: "JSClosure.release() is no longer necessary")
```

Sources/_CJavaScriptKit/_CJavaScriptKit.c
```diff
@@ -35,8 +35,15 @@ void swjs_cleanup_host_function_call(void *argv_buffer) {
 /// Notes: If you change any interface of runtime library, please increment
 /// this and `SwiftRuntime.version` in `./Runtime/src/index.ts`.
 __attribute__((export_name("swjs_library_version")))
-int swjs_library_version() {
+int swjs_library_version(void) {
     return 701;
 }
 
+int _library_features(void);
```
j-f1, line 42: Does this addition warrant a version bump?

kateinoigakukun: Good catch!

Runtime/src/index.ts
kateinoigakukun: I forgot to apply prettier 😅

Runtime/src/index.ts
kateinoigakukun: That's a cache of object and it should not be touched by other than the `closureHeap` getter.

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
kateinoigakukun: I think that behavior seems natural.  In what use cases is that behavior useful?

j-f1: Adding multiple event listeners to a DOM node, for example.

kateinoigakukun: In that case, the closure instances are retained by host environment, so the Swift instances are also retained and they should have different identifies at a time

j-f1: The closure itself (`{ _ in ... }`) is retained by the host environment, but the `JSClosure` wrapper is immediately deallocated because it is not retained by anything.

kateinoigakukun: Ah, I see. So we should retain the JSClosure instance itself.

Runtime/src/index.ts
j-f1: That makes sense!

j-f1 (approved):

PR #142, opened 2021-11-09
Fix recursion in TypedArray initializer

PatrickPijnappel:
The original unconstrained version would cause infinite recursion.


MaxDesiatov (approved):
I've added a simple test, which I hope is enough, but let me know it doesn't reflect the code you've used to reproduce this on your side.

PatrickPijnappel:
@MaxDesiatov that looks good! Sorry I didn't get to this earlier had been really busy.

MaxDesiatov:
No problem, thanks again for the PR!

PR #143, opened 2021-11-21
Bump to v0.11.0, update `CHANGELOG.md` & `Example`

MaxDesiatov:
I had to remove SwiftWasm 5.3 and 5.4 from CI because https://github.com/OpenCombine/OpenCombine/issues/225 is reproducible with all code, not just OpenCombine, with old toolchains when Xcode 13.1 is installed.

I've also updated `Example` code to try using `fetch` with `async`/`await` and a section with a simple guide for this to `README.md`

Node.js and npm dependencies together with webpack.js config in `Example` directory were also updated.


MaxDesiatov:
I think it's a bit of a pain to test Tokamak with this new JSKit version without releasing everything, so I think I'm going to do it like this:

1. Release JSKit 0.11.0.
2. Release `carton` 0.12.0 with new JSKit runtime.
3. Release new Tokamak 0.9.0 with basic `async`/`await` support and other new features if all works fine.
4. Release SwiftWasm 5.5 if all of the above seems to work fine.

WDYT?

j-f1 (approved):

PR #144, opened 2021-11-22
Remove macOS Monterey requirement from `Package.swift`

MaxDesiatov:
This currently prevents Tokamak from being built on macOS 11 or earlier macOS versions. I think the original concern is no longer an issue, where back-deployment with Xcode 13.2 to old macOS versions will become available really soon. In fact, beta versions of Xcode 13.2 are already available on GitHub Actions if we need to test with those.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,12 +4,6 @@ import PackageDescription
 
 let package = Package(
     name: "JavaScriptKit",
-    platforms: [
-        // This package doesn't work on macOS host, but should be able to be built for it
-        // for developing on Xcode. This minimum version requirement is to prevent availability
-        // errors for Concurrency API, whose runtime support is shipped from macOS 12.0
-        .macOS("12.0")
-    ],
     products: [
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
In theory, SwiftPM should ignore that requirement when building for macOS because JavaScriptKit is declared as an optional dependency in Tokamak only for wasm32-unknown-wasi. 
But for now, it's ok to remove the requirement. We need to describe that users need to add `.unsafeFlags(["-Xfrontend", "-disable-availability-checking"])` in their Package.swift to develop on Xcode under macOS 11.

MaxDesiatov:
@kateinoigakukun Would mentioning that Xcode 13.2 is required for development of macOS apps depending on JavaScriptKit be enough?

MaxDesiatov:
Great, I've added a clarification to `README.md`. Also added full concurrency example there to make it more obvious why we have these weird Xcode/macOS requirements 😅

kateinoigakukun (approved):

PR #147, opened 2022-01-03
Return from runtime functions instead of taking a pointer where possible

j-f1:


kateinoigakukun (approved):
Could you bump up library version in swjs_library_version because this change breaks runtime library compatibility?

j-f1:
Done!

PR #148, opened 2022-01-03
proposal: mechanism for using alternate backends

j-f1:
To use this API, first define a enum or struct conforming to `JSBridge` and define the required methods. Then, call `JSObject.global(using: MyJSBridge.self)` to get the global object.

Currently, `JSTypedArray` and `JSBridgedClass` only work with the default bridge, but I didn’t want to put too much effort into figuring out how to get those APIs to work properly if this approach didn’t end up getting accepted.

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -151,8 +151,8 @@ export class SwiftRuntime {
 
     setInstance(instance: WebAssembly.Instance) {
         this.instance = instance;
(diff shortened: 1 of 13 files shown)


j-f1:
So it looks like it is slower. Tragic.

j-f1:
@kateinoigakukun’s proposal in discord is much better. Closing this for now.

PR #150, opened 2022-01-08
Split runtime into multiple files

j-f1:
This PR:

- splits the runtime TypeScript code up into several separate modules
- adds Rollup to compile that back into a single JS file
- uses Rollup to produce two bundles instead of one:
  - one that uses UMD to allow use with Node-style `require`, or by embedding it with a `<script>` tag and using `window.JavaScriptKit`
  - one that uses ES Modules
- refactor the codebase to use a `Memory` class that provides easy access to memory and the heap
- check for FinalizationRegistry availability in `SwiftClosureDeallocator` constructor rather than outside of it

### DIFF
--- a/Runtime/src/closure-heap.ts
+++ b/Runtime/src/closure-heap.ts
@@ -1,10 +1,10 @@
-import { SwiftRuntimeExportedFunctions } from "./types";
+import { ExportedFunctions } from "./types";
 
 /// Memory lifetime of closures in Swift are managed by Swift side
 export class SwiftClosureDeallocator {
     private functionRegistry: FinalizationRegistry<number>;
 
-    constructor(exports: SwiftRuntimeExportedFunctions) {
+    constructor(exports: ExportedFunctions) {
         if (typeof FinalizationRegistry === "undefined") {
             throw new Error(
                 "The Swift part of JavaScriptKit was configured to require the availability of JavaScript WeakRefs. Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable features that use WeakRefs."
--- a/Runtime/src/closure-heap.ts
+++ b/Runtime/src/closure-heap.ts
@@ -1,0 +1,16 @@
+import { SwiftRuntimeExportedFunctions } from "./types";
+
+/// Memory lifetime of closures in Swift are managed by Swift side
+export class SwiftClosureHeap {
+    private functionRegistry: FinalizationRegistry<number>;
+
+    constructor(exports: SwiftRuntimeExportedFunctions) {
+        this.functionRegistry = new FinalizationRegistry((id) => {
+            exports.swjs_free_host_function(id);
+        });
+    }
+
+    alloc(func: any, func_ref: number) {
+        this.functionRegistry.register(func, func_ref);
+    }
+}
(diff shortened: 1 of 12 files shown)


kateinoigakukun:
Some tests are failing 👀 

```console

Error in Literal Conversion

MessageError(message: "Expect to be equal \"212205744152385\" and \"4746794007244308480\"", file: "/home/runner/work/JavaScriptKit/JavaScriptKit/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift", line: 26, column: 28)

Swift/ErrorType.swift:200: Fatal error: Error raised at top level: PrimaryTests.MessageError(message: "Expect to be equal \"212205744152385\" and \"4746794007244308480\"", file: "/home/runner/work/JavaScriptKit/JavaScriptKit/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift", line: 26, column: 28)

RuntimeError: unreachable
    at $ss17_assertionFailure__4file4line5flagss5NeverOs12StaticStringV_SSAHSus6UInt32VtF (wasm://wasm/02991d2a:wasm-function[2633]:0xccfd3)
    at swift_errorInMain (wasm://wasm/02991d2a:wasm-function[4819]:0x14b4ac)
    at main (wasm://wasm/02991d2a:wasm-function[1532]:0x76025)
    at main (wasm://wasm/02991d2a:wasm-function[29674]:0x622f25)
    at __main_void (wasm://wasm/02991d2a:wasm-function[29675]:0x622fa1)
    at __original_main (wasm://wasm/02991d2a:wasm-function[29672]:0x622ee6)
    at _start (wasm://wasm/02991d2a:wasm-function[45]:0x10884)
    at d.start (/home/runner/work/JavaScriptKit/JavaScriptKit/IntegrationTests/node_modules/@wasmer/wasi/lib/index.cjs.js:77:280)
    at startWasiTask (/home/runner/work/JavaScriptKit/JavaScriptKit/IntegrationTests/lib.js:47:10)
```

j-f1:
It turned out to be an endianness issue, oof. Hoping this change fixes it 🤞

Runtime/src/closure-heap.ts
```diff
@@ -0,0 +1,22 @@
+import { ExportedFunctions } from "./types";
+
+/// Memory lifetime of closures in Swift are managed by Swift side
+export class SwiftClosureDeallocator {
+    private functionRegistry: FinalizationRegistry<number>;
+
+    constructor(exports: ExportedFunctions) {
+        if (typeof FinalizationRegistry === "undefined") {
+            throw new Error(
+                "The Swift part of JavaScriptKit was configured to require the availability of JavaScript WeakRefs. Please build with `-Xswiftc -DJAVASCRIPTKIT_WITHOUT_WEAKREFS` to disable features that use WeakRefs."
```
MaxDesiatov, line 10: nit: can we make this a multi-line string? No need to display it as multi-line to the user, just a matter of formatting in the source code.

Runtime/src/types.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export type TypedArray =
+    | Int8ArrayConstructor
+    | Uint8ArrayConstructor
+    | Int16ArrayConstructor
+    | Uint16ArrayConstructor
+    | Int32ArrayConstructor
+    | Uint32ArrayConstructor
+    // | BigInt64ArrayConstructor
+    // | BigUint64ArrayConstructor
```
MaxDesiatov, line 102: nit: needs a clarifying comment explaining why these are commented out

MaxDesiatov (approved):
Overall seems legit 👍 but I'd wait for at least one more approval from the team

kateinoigakukun (approved):

PR #153, opened 2022-01-25
Fix detached ArrayBuffer errors

yonihemi:
Fixes a regression caused by #150: 
`Unhandled Promise Rejection: TypeError: Underlying ArrayBuffer has been detached from the view` (Safari)
`TypeError: Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer` (Chrome)

It's troubling our unit test don't catch this but with my limited knowledge of JS memory management I don't know how to reliably recreate it.

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -177,7 +177,7 @@ export class SwiftRuntime {
         },
 
         swjs_decode_string: (bytes_ptr: pointer, length: number) => {
-            const bytes = this.memory.bytes.subarray(
+            const bytes = this.memory.bytes().subarray(
                 bytes_ptr,
                 bytes_ptr + length
             );
--- a/Runtime/src/memory.ts
+++ b/Runtime/src/memory.ts
@@ -3,29 +3,27 @@ import { pointer } from "./types";
 
 export class Memory {
     readonly rawMemory: WebAssembly.Memory;
-    readonly bytes: Uint8Array;
-    readonly dataView: DataView;
 
     private readonly heap = new SwiftRuntimeHeap();
 
     constructor(exports: WebAssembly.Exports) {
         this.rawMemory = exports.memory as WebAssembly.Memory;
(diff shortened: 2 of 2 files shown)


kateinoigakukun:
Good catch! As far as I understand, `WebAssembly.Memory` detaches its internal ArrayBuffer when `memory.grow` instruction is executed because `ArrayBuffer` is a fixed-size array. So we need to access `WebAssembly.Memory.buffer` every time.

kateinoigakukun (approved):

kateinoigakukun:
Could you add a test case to kill this bug by using `__builtin_wasm_memory_grow`?

yonihemi:
> Good catch! As far as I understand, `WebAssembly.Memory` detaches its internal ArrayBuffer when `memory.grow` instruction is executed because `ArrayBuffer` is a fixed-size array. So we need to access `WebAssembly.Memory.buffer` every time.

In my case there is no explicit call to `.grow` anywhere, perhaps it's detached in more cases.

tierracero:
hey hi  :) 

hope your all doing fine 

seems I got the same error 

JavascriptKit 0.17.0

> app.js:139 TypeError: Cannot perform DataView.prototype.setBigUint64 on a detached ArrayBuffer
>     at DataView.setBigUint64 (<anonymous>)
>     at clock_res_get (index.esm.js:114:167)
>     at __wasi_clock_res_get (1ddb7b2a:0x3b79363)
>     at clock_getres (1ddb7b2a:0x3b767a8)
>     at __CFDateInitialize (1ddb7b2a:0x2ff76e7)
>     at __CFInitialize (1ddb7b2a:0x2fb8851)
>     at __wasm_call_ctors (1ddb7b2a:0x76f37)
>     at _initialize (1ddb7b2a:0x76f56)
>     at startWasiTask (app.js:133:26)

tierracero:
sure give a min, thank you.

PR #159, opened 2022-02-03
Add async closure support

### LINKED ISSUE #156: JSPromise(resolver:) usage
Could somebody provide an example of how to use `JSPromise(resolver:)`? I'm having hard time decoding the following comment in the source code:
> Creates a new `JSPromise` instance from a given `resolver` closure. `resolver` takes two closure that your code should call to either resolve or reject this `JSPromise` instance.

### LINKED ISSUE #157: Asynchronous calls in JSClosure
What is the best way to handle calls to asynchronous code inside a `JSClosure`, especially if one needs to return the result of such call from the `JSClosure`? Consider the following example:
```swift
let closure = .object(JSClosure { (arguments: [JSValue]) in
    guard let url = arguments.first?.string else {
        return JSValue.undefined
    }
    return await fetch(url)
})
```
This results in the following error:
> cannot pass function of type '([JSValue]) async -> JSValue' to parameter expecting synchronous function type

The reason is obvious, but is there a nice way to work around this?

j-f1:
Fixes #157, fixes #156

- [x] Tests

### DIFF
--- a/IntegrationTests/TestSuites/Sources/ConcurrencyTests/UnitTestUtils.swift
+++ b/IntegrationTests/TestSuites/Sources/ConcurrencyTests/UnitTestUtils.swift
@@ -41,6 +41,18 @@ struct MessageError: Error {
     }
 }
 
+func expectGTE<T: Comparable>(
+    _ lhs: T, _ rhs: T,
+    file: StaticString = #file, line: UInt = #line, column: UInt = #column
+) throws {
+    if lhs < rhs {
+        throw MessageError(
+            "Expected \(lhs) to be greater than or equal to \(rhs)",
+            file: file, line: line, column: column
+        )
+    }
+}
+
 func expectEqual<T: Equatable>(
     _ lhs: T, _ rhs: T,
     file: StaticString = #file, line: UInt = #line, column: UInt = #column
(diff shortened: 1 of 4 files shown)


j-f1:
Currently, I think all it’s missing are tests for the new functionality. I’m not sure when I will get around to adding those, but if you would like to open a PR against this PR with tests, I’d be happy to review and merge it in.

j-f1:
Currently, I think all it’s missing are tests for the new functionality. I’m not sure when I will get around to adding those, but if you would like to open a PR against this PR with tests, I’d be happy to review and merge it in.

j-f1:
There’s a test, but most of the new API surface is untested. I’m happy to merge if that’s ok with you, but otherwise I don’t really have any ideas for more tests.

MaxDesiatov:
@kateinoigakukun would you prefer more test coverage here? If so, how would you cover those missing areas?

kateinoigakukun:
I want more tests around the use of `catch` and `then(success:failure)` APIs at least to avoid unintentional API breakage in the future.

MaxDesiatov (approved):
LGTM, thanks!

kateinoigakukun (approved):

PR #173, opened 2022-03-30
Improve error messages when JS code throws exceptions

pedrovgs:
We've noticed when JS code throws an exception evaluated from ``JSFunction`` the error reported shows this message:

```
RuntimeError: Unreachable code should not be executed (evaluating 'this.exports.swjs_call_host_function(e,a,r,u)')
  [native code] at line 0
   ../node_modules/javascript-kit-swift/Runtime/lib/index.mjs at line 348
```

After reviewing this with the team, @kateinoigakukun suggested there where a try/catch capturing the exception we should remove in order to improve this scenario. 

This PR adds two new unsafe versions of already existing functions where the JS side of the project doesn't capture any exception so the final result is a better error reporting like this:

Screen_Shot_2022-03-30_at_10 35 52

### DIFF
--- a/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
+++ b/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
@@ -700,6 +700,59 @@ try test("Exception") {
     // }
     // ```
     //
+    let globalObject1 = JSObject.global.globalObject1
+    let prop_9: JSValue = globalObject1.prop_9
+
+    // MARK: Throwing method calls
+    let error1 = try expectThrow(try prop_9.object!.throwing.func1!())
+    try expectEqual(error1 is JSValue, true)
+    let errorObject = JSError(from: error1 as! JSValue)
+    try expectNotNil(errorObject)
+
+    let error2 = try expectThrow(try prop_9.object!.throwing.func2!())
+    try expectEqual(error2 is JSValue, true)
+    let errorString = try expectString(error2 as! JSValue)
+    try expectEqual(errorString, "String Error")
+
+    let error3 = try expectThrow(try prop_9.object!.throwing.func3!())
+    try expectEqual(error3 is JSValue, true)
+    let errorNumber = try expectNumber(error3 as! JSValue)
+    try expectEqual(errorNumber, 3.0)
+
+    // MARK: Simple function calls
+    let error4 = try expectThrow(try prop_9.func1.function!.throws())
+    try expectEqual(error4 is JSValue, true)
+    let errorObject2 = JSError(from: error4 as! JSValue)
+    try expectNotNil(errorObject2)
+
+    // MARK: Throwing constructor call
+    let Animal = JSObject.global.Animal.function!
+    _ = try Animal.throws.new("Tama", 3, true)
+    let ageError = try expectThrow(try Animal.throws.new("Tama", -3, true))
+    try expectEqual(ageError is JSValue, true)
+    let errorObject3 = JSError(from: ageError as! JSValue)
+    try expectNotNil(errorObject3)
+}
+
+try test("Unhandled Exception") {
+    // ```js
+    // global.globalObject1 = {
+    //   ...
+    //   prop_9: {
+    //       func1: function () {
+    //           throw new Error();
+    //       },
+    //       func2: function () {
+    //           throw "String Error";
+    //       },
+    //       func3: function () {
+    //           throw 3.0
+    //       },
+    //   },
+    //   ...
+    // }
+    // ```
+    //
 
     let globalObject1 = JSObject.global.globalObject1
     let prop_9: JSValue = globalObject1.prop_9
@@ -719,14 +772,6 @@ try test("Exception") {
     try expectEqual(error3 is JSValue, true)
     let errorNumber = try expectNumber(error3 as! JSValue)
     try expectEqual(errorNumber, 3.0)
-
-    // MARK: Throwing constructor call
-    let Animal = JSObject.global.Animal.function!
-    _ = try Animal.throws.new("Tama", 3, true)
-    let ageError = try expectThrow(try Animal.throws.new("Tama", -3, true))
-    try expectEqual(ageError is JSValue, true)
-    let errorObject3 = JSError(from: ageError as! JSValue)
-    try expectNotNil(errorObject3)
 }
 
 /// If WebAssembly.Memory is not accessed correctly (i.e. creating a new view each time),
--- a/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
+++ b/IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
@@ -705,26 +705,20 @@ try test("Exception") {
     let prop_9: JSValue = globalObject1.prop_9
 
     // MARK: Throwing method calls
-    let error1 = try expectThrow(try prop_9.object!.throwing.func1!())
+    let error1 = try wrapUnsafeThrowableFunction { _ = prop_9.object!.func1!() }
     try expectEqual(error1 is JSValue, true)
     let errorObject = JSError(from: error1 as! JSValue)
     try expectNotNil(errorObject)
 
-    let error2 = try expectThrow(try prop_9.object!.throwing.func2!())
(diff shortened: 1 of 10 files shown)


pedrovgs:
Hey @kateinoigakukun thank you so much for your help. I applied the changes you suggested but I've found one test I can't fix related to JSClosure. ``test("Closure Lifetime")`` seems to be failing but I don't get what's going on. Could you help me with this? Thank you in advance 😃

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -93,11 +93,11 @@ internal func invokeJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJS
             var payload1 = JavaScriptPayload1()
             var payload2 = JavaScriptPayload2()
             if let thisId = this?.id {
-                _call_function_with_this(thisId,
+                _call_function_with_this_unsafe(thisId,
```
j-f1, line 96: `invokeJSFunction` is also used by `JSThrowingFunction`, so I think you’d have to copy-paste the code and update it as appropriate.

I think the non-throwing version should do an `assert(!kindAndFlags.isException` before simply returning the `JSValue`:

```swift
internal func invokeNonThrowingJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJSValue], this: JSObject?) throws -> JSValue {
    rawValues.withUnsafeBufferPointer { bufferPointer -> (JSValue, Bool) in
        let argv = bufferPointer.baseAddress
        let argc = bufferPointer.count
        var kindAndFlags = JavaScriptValueKindAndFlags()
        var payload1 = JavaScriptPayload1()
        var payload2 = JavaScriptPayload2()
        if let thisId = this?.id {
            _call_function_with_this_unsafe(
                thisId,
                jsFunc.id, argv, Int32(argc),
                &kindAndFlags, &payload1, &payload2
            )
        } else {
            _call_function_unsafe(
                jsFunc.id, argv, Int32(argc),
                &kindAndFlags, &payload1, &payload2
            )
        }
        assert(!kindAndFlags.isException)
        let result = RawJSValue(kind: kindAndFlags.kind, payload1: payload1, payload2: payload2)
        return result.jsValue()
    }
}

File: Runtime/src/index.ts [hunk 2]
```diff
@@ -226,6 +226,29 @@ export class SwiftRuntime {
...
+                JSValue.decodeArray(argv, argc, this.memory)
+            );
+            JSValue.write(
+                result,
+                kind_ptr,
+                payload1_ptr,
+                payload2_ptr,
+                false,
+                this.memory
+            );
```
j-f1, line 250: This will tell the Wasm side if there was an unexpected JS-side exception

```suggestion
            let isException = true;
            try {
                const result = Reflect.apply(
                    func,
                    undefined,
                    JSValue.decodeArray(argv, argc, this.memory)
                );
                isException = false;
                JSValue.write(
                    result,
                    kind_ptr,
                    payload1_ptr,
                    payload2_ptr,
                    false,
                    this.memory
                );
            } finally {
                if (isException) {
                    JSValue.write(
                        undefined,
                        kind_ptr,
                        payload1_ptr,
                        payload2_ptr,
                        true,
                        this.memory
                    );
                }
            }
```

pedrovgs:
Thanks for the tip @j-f1 I didn't notice the code was being invoked from ``JSThrowingFunction``. I've added a new commit with the suggested change 😃

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
pedrovgs: Thanks! I've applied the change already in a new commit 😃

Runtime/src/index.ts
pedrovgs: Thanks @j-f1 I'm going to update the other function I created with a similar approach as well!

kateinoigakukun: "an unexpected JS-side exception" means exceptions thrown by our runtime, right?

IntegrationTests/TestSuites/Sources/PrimaryTests/main.swift
```diff
@@ -705,27 +705,21 @@ try test("Exception") {
     let prop_9: JSValue = globalObject1.prop_9
 
     // MARK: Throwing method calls
-    let error1 = try expectThrow(try prop_9.object!.throwing.func1!())
+    let error1 = try wrapUnsafeThrowableFunction { _ = prop_9.object!.func1!() }
```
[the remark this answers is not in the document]
pedrovgs: Done! I've created a new test case for the new behavior and maintained the previous coverage as is. Thanks!

Runtime/src/index.ts
j-f1: @kateinoigakukun Not quite! This is for when the user calls a function on the JS side that throws an exception:

```swift
JSObject.global.atob.function!(7)
```

This change will allow the exception to bubble up to the devtools console (along with halting Swift execution), making it easier to debug.

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -93,15 +93,16 @@ internal func invokeJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJS
...
-                _call_function(
+                _call_function_unsafe(
                     jsFunc.id, argv, Int32(argc),
                     &kindAndFlags, &payload1, &payload2
                 )
             }
+            assert(!kindAndFlags.isException)
             let result = RawJSValue(kind: kindAndFlags.kind, payload1: payload1, payload2: payload2)
             return (result.jsValue(), kindAndFlags.isException)
         }
```
j-f1, line 108: one small change here — you don’t need to handle the exception case once you’ve `assert()`-ed it away:

```swift
private func invokeNonThrowingJSFunction(...) throws -> JSValue {
    arguments.withRawJSValues { rawValues in
        rawValues.withUnsafeBufferPointer { bufferPointer -> JSValue in
            // ...
            assert(!kindAndFlags.isException)
            let result = RawJSValue(kind: kindAndFlags.kind, payload1: payload1, payload2: payload2)
            return result.jsValue()
        }
    }
}
```

Runtime/src/index.ts
```diff
@@ -226,6 +226,44 @@ export class SwiftRuntime {
                 this.memory
             );
         },
+        swjs_call_function_unsafe: (
```
j-f1, line 229: Also, maybe call these `swjs_call_function_no_catch` to more clearly say what they do? Since they don’t exhibit undefined behavior or cause memory safety issues, I think they need to be called “unsafe.” But I’m definitely not 100% confident here and would like to hear your thoughts!

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
kateinoigakukun: Also `throws` is not required :)

pedrovgs: Yeap you are absolutely right! I've applied your suggestions in a new commit 😃

Runtime/src/index.ts
pedrovgs: I like your idea! I've added a new commit renaming `unsafe` to `no_catch` 😃

pedrovgs:
@j-f1 @kateinoigakukun thank you all for your comments and your review. I think I've applied all your suggestions. Please let me know if you want me to review any other detail or change anything 😃  Thanks!

kateinoigakukun (commented):
Sorry, one more thing. Could you bump up the interface version in
- https://github.com/swiftwasm/JavaScriptKit/blob/b6b7b981b994e2afba305cbee8c8de21ff0dd942/Runtime/src/index.ts#L17
- https://github.com/swiftwasm/JavaScriptKit/blob/b6b7b981b994e2afba305cbee8c8de21ff0dd942/Sources/_CJavaScriptKit/_CJavaScriptKit.c#L39

pedrovgs:
No need to be sorry @j-f1 😃 Version value updated from 705 to 706!

kateinoigakukun (approved):

kateinoigakukun:
Thank you very much!

PR #174, opened 2022-03-30
Updates for DOMKit

j-f1:
- [breaking] The `ConvertibleToJSValue` and `JSBridgedType` protocols have changed:
  - To conform to `ConvertibleToJSValue`, you must now provide a `jsValue: JSValue { get }` property instead of a `jsValue() -> JSValue` method.
    - existing code that uses `ConvertibleToJSValue` will still work, as there is a protocol extension that provides a deprecated `jsValue()` method.
  - `JSBridgedType` has dropped its `value` property requirement in favor of inheriting the `jsValue` property from `ConvertibleToJSValue`. To update your conforming types, simply rename `value` to `jsValue`.

- [breaking] The `ConvertibleToJSValue` conformance on `Array` and `Dictionary` has been swapped from the `== ConvertibleToJSValue` case to the `: ConvertibleToJSValue` case.
  - This means that e.g. `[String]` is now `ConvertibleToJSValue`, but `[ConvertibleToJSValue]` no longer conforms
  - the `jsValue()` method still works in both cases
  - to adapt existing code, use one of these approaches:
    - use generics where possible (for single-type arrays)
    - call `.map { $0.jsValue() }` (or `mapValues`) to get an array/dictionary of `JSValue` which you can then use as `ConvertibleToJSValue`
    - add `.jsValue` to the end of all of the values in the array/dictionary literal

- Allow people to do `object[JSString(...)]!(args...)`
  - it was previously only possible to do this with `String` keys, even though the other property accessors worked with `JSString`s too.

- Lower the Xcode concurrency requirements so they work with the new backporting features

- Implement `JSUInt8ClampedArray` (wrapping `Uint8ClampedArray`)
  - now that I think about it, since Swift defines overflow behaviors for all numbers, this type doesn’t make much sense — the only `DOMKit` usage is in [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData), so perhaps it would be better to alias it to `JSTypedArray<UInt8>`?
  - …but they’re two different classes in JS land so I think it’s useful to have both

- ~Allow users to subclass JSObject~ (removed)
  - this is useful for providing ergonomic access to struct-like objects (i.e. options and API return values)
  - now that I think about it, this whole area is full of pitfalls and may be a bad approach (i.e. `fromJSValue` won’t work)
    - will try to re-implement my needs in userland and then remove this capability

- ~Add a new `JSObject(cloning: JSObject)` initializer, enabling people to point two JSObject instances at the same object.~ (removed)
  - this is necessary to construct new DOMKit dictionary objects, as there is currently no valid way to call `super` in a `JSObject` subclass, and there is unfortunately no way to “transmute” a JSObject into a subclass after getting it from 
  - ~this requires a new `swjs_retain` call to keep reference counts correct~ (removed)

- ~[breaking] rename the `JSPromise/value` API to `JSPromise/get()` to make it actually possible to use~ (reverted)
  - currently, it’s shadowed by `JSBridgedType.value`
  - the API is inspired by `Result/get()`
  - an alternative would be to rename the `JSBridgedType` requirement to `jsValue`, but this would make autocomplete for that somewhat annoying.
    - a follow-on change from this would be to rename `ConvertibleToJSValue/jsValue()` to `toJSValue()` to match `ConstructibleFromJSValue`’s `JSValue/fromJSValue()` method.
  - thoughts?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,9 +8,10 @@ jobs:
     name: Build and Test
     strategy:
       matrix:
-        os: [macos-10.15, macos-11, ubuntu-18.04, ubuntu-20.04]
+        os: [macos-10.15, macos-11, macos-12, ubuntu-18.04, ubuntu-20.04]
         toolchain:
           - wasm-5.5.0-RELEASE
+          - wasm-5.6.0-RELEASE
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
@@ -29,9 +30,18 @@ jobs:
   native-build:
     # Check native build to make it easy to develop applications by Xcode
     name: Build for native target
-    runs-on: macos-11
+    strategy:
+      matrix:
+        include:
+          - os: macos-10.15
+            xcode: Xcode_12.4
+          - os: macos-11
+            xcode: Xcode_13.2.1
+          - os: macos-12
+            xcode: Xcode_13.3
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
       - run: swift build
         env:
-          DEVELOPER_DIR: /Applications/Xcode_13.2.1.app/Contents/Developer/
+          DEVELOPER_DIR: /Applications/${{ matrix.xcode }}.app/Contents/Developer/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+# 0.13.0 (31 March 2022)
+
+This release improves handling of JavaScript exceptions and compatibility with Xcode.
+
+Thanks to [@kateinoigakukun](https://github.com/kateinoigakukun), [@pedrovgs](https://github.com/pedrovgs), and
+[@valeriyvan](https://github.com/valeriyvan) for contributions!
+
+**Closed issues:**
+
+- UserAgent support? ([#169](https://github.com/swiftwasm/JavaScriptKit/issues/169))
+- Compile error on macOS 12.2.1 ([#167](https://github.com/swiftwasm/JavaScriptKit/issues/167))
+
+**Merged pull requests:**
+
+- Improve error messages when JS code throws exceptions ([#173](https://github.com/swiftwasm/JavaScriptKit/pull/173)) via [@pedrovgs](https://github.com/pedrovgs)
+- Update npm dependencies ([#175](https://github.com/swiftwasm/JavaScriptKit/pull/175)) via [@MaxDesiatov](https://github.com/MaxDesiatov)
+- Bump minimist from 1.2.5 to 1.2.6 in /Example ([#172](https://github.com/swiftwasm/JavaScriptKit/pull/172)) via [@dependabot[bot]](https://github.com/dependabot[bot])
+- Use availability guarded APIs under @available for Xcode development ([#171](https://github.com/swiftwasm/JavaScriptKit/pull/171)) via [@kateinoigakukun](https://github.com/kateinoigakukun)
+- Fix warning in snippet ([#166](https://github.com/swiftwasm/JavaScriptKit/pull/166)) via [@valeriyvan](https://github.com/valeriyvan)
+- Bump follow-redirects from 1.14.5 to 1.14.8 in /Example ([#165](https://github.com/swiftwasm/JavaScriptKit/pull/165)) via [@dependabot[bot]](https://github.com/dependabot[bot])
+
 # 0.12.0 (08 February 2022)
 
 This release introduces a [major refactor](https://github.com/swiftwasm/JavaScriptKit/pull/150) of the JavaScript runtime by [@j-f1] and several performance enhancements. 
--- a/Example/package-lock.json
+++ b/Example/package-lock.json
@@ -14,9 +14,9 @@
         "javascript-kit-swift": "file:.."
       },
       "devDependencies": {
-        "webpack": "^5.64.2",
-        "webpack-cli": "^4.9.1",
-        "webpack-dev-server": "^4.5.0"
+        "webpack": "^5.70.0",
+        "webpack-cli": "^4.9.2",
+        "webpack-dev-server": "^4.7.4"
       }
     },
     "..": {
@@ -24,11 +24,11 @@
       "version": "0.12.0",
       "license": "MIT",
       "devDependencies": {
-        "@rollup/plugin-typescript": "^8.3.0",
-        "prettier": "2.5.1",
-        "rollup": "^2.63.0",
+        "@rollup/plugin-typescript": "^8.3.1",
+        "prettier": "2.6.1",
+        "rollup": "^2.70.0",
         "tslib": "^2.3.1",
-        "typescript": "^4.5.5"
+        "typescript": "^4.6.3"
       }
     },
     "../node_modules/prettier": {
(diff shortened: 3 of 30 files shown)


j-f1:
An update on the JSObject changes (see https://github.com/swiftwasm/DOMKit/pull/10): I no longer think the subclassing capability (and therefore the `init(cloning:)` API) is a good idea, so I will remove that feature from this PR and rewrite DOMKit to use a standard `JSBridgedType` instead.

kateinoigakukun:
> [breaking] rename the JSPromise/value API to JSPromise/get() to make it actually possible to use

I have some objections to this. `Task` uses `Task.value` and also [`Result` is going to use `Result.value` instead of `Result.get()`](https://forums.swift.org/t/add-value-property-to-result/50253/), so I like `value` for consistency.

> [breaking] swap the ConvertibleToJSValue conformance on Array and Dictionary to the : ConvertibleToJSValue case instead of the == ConvertibleToJSValue case.

I don't understand the necessity of this change, could you elaborate on the reason? `[String]` is already `ConvertibleToJSValue` without this change.
My intention for the conformance requirement was to conform both `[X] where X: ConvertibleToJSValue` and `[ConvertibleToJSValue]` to `ConvertibleToJSValue`. Think the following case.

```swift
protocol P0 {}

extension Array: P0 where Element == P0 {}
extension Dictionary: P0 where Value == P0, Key == String {}
// If forcing `Element` or `Value` to be non-existential, it breaks existential cases
// extension Array: P0 where Element: P0 {}
// extension Dictionary: P0 where Value: P0, Key == String {}

func takeP0<X: P0>(_: X) {}

func giveExistentialArray(_ xs: [P0]) {
  takeP0(xs)
}
func giveGenericArray<X: P0>(_ xs: [X]) {
  takeP0(xs)
}

func giveExistentialDictionary(_ xs: [String: P0]) {
  takeP0(xs)
}
func giveGenericArray<X: P0>(_ xs: [String: X]) {
  takeP0(xs)
}
```

> Implement JSUInt8ClampedArray (wrapping Uint8ClampedArray)

As far as I understand, `Uint8ClampedArray` has a setter subscript with `Double` and getter subscript with `UInt8`. But such subscript cannot be representable in Swift, so limiting the setter with `UInt8` seems reasonable to me.

j-f1:
> I have some objections to this. `Task` uses `Task.value` and also [`Result` is going to use `Result.value` instead of `Result.get()`](https://forums.swift.org/t/add-value-property-to-result/50253/), so I like `value` for consistency.

Oh cool, I didn’t know about that. We should definitely use `value`, then. That would necessitate changing the `JSBridgedType` API and possibly the `ConvertibleToJSValue` API — how does this look?

```swift
public protocol JSBridgedType: JSValueCompatible, CustomStringConvertible {
    // renamed from `value`
    var jsValue: JSValue { get }
    init?(from jsValue: JSValue)
}

public protocol ConvertibleToJSValue {
    // renamed from `jsValue` to avoid an autocomplete conflict
    func toJSValue() -> JSValue
}
```

Alternatively, we could have a unified API:

```swift
public protocol JSBridgedType: JSValueCompatible, CustomStringConvertible {
    // [the jsValue member is inherited from ConvertibleToJSValue]
    init?(from jsValue: JSValue)
}

public protocol ConvertibleToJSValue {
    var jsValue: JSValue { get }
}
```

Note that:

- `ConstructibleFromJSValue` cannot use an initializer because we have to return already-existing instances of e.g. booleans from inside of the `JSValue`
- we could have both `var jsValue: JSValue` and `func jsValue() -> JSValue`, but Xcode doesn’t autocomplete it properly, and it is confusing IMO


> I don't understand the necessity of this change, could you elaborate on the reason? `[String]` is already `ConvertibleToJSValue` without this change.

Unfortunately, it isn’t. While you can call `[String]().jsValue()`, it isn’t formally part of the protocol. Swift only allows a type (such as `Array`) to conform to a protocol once, so it must be either `[ConvertibleToJSValue]` or `Array where Element: ConvertibleToJSValue` that conforms. You can still call `.jsValue()` on the other one, but it can’t be used as as a parameter to a function that takes any `T: ConvertibleToJSValue`, for example. My hope is that anyone who needs to pass a `[ConvertibleToJSValue]` to a parameter of type `ConvertibleToJSValue` will be able to call `.jsValue()` on it, to get back a `JSValue` which is `ConvertibleToJSValue`.

… sorry that was kinda badly written but hopefully it gets my point across!

I only made this change because my build of DOMKit was breaking.

> Is this feature necessary only if we will take subclassing approach? [referring to `JSObject(cloning: JSObject)`]

Yep, and I’ve pushed an update to remove it.

> As far as I understand, `Uint8ClampedArray` has a setter subscript with `Double` and getter subscript with `UInt8`. But such subscript cannot be representable in Swift, so limiting the setter with `UInt8` seems reasonable to me.

👍

kateinoigakukun:
@j-f1 

> ```swift
> public protocol JSBridgedType: JSValueCompatible, CustomStringConvertible {
>     // renamed from `value`
>     var jsValue: JSValue { get }
>     init?(from jsValue: JSValue)
> }
> 
> public protocol ConvertibleToJSValue {
>     // renamed from `jsValue` to avoid an autocomplete conflict
>     func toJSValue() -> JSValue
> }
> ```

It looks like fine to me 👍 


> Unfortunately, it isn’t. While you can call [String]().jsValue(), it isn’t formally part of the protocol.

Yes, `[String]` is not `ConvertibleToJSValue` but it's implicitly convertible to `ConvertibleToJSValue` by type-coercing because Array and Dictionary are special container types that are "covariant".
Could you tell me the real breaking case in DOMKit?

j-f1:
> Could you tell me the real breaking case in DOMKit?

The issue was that there’s a property wrapper for JS object keys:

```swift
@propertyWrapper public struct ReadWriteAttribute<Wrapped: JSValueCompatible> {
  // ...
}

class Foo: JSValueCompatible {
    // ...
    @ReadWriteAttribute var bar: [String]
}
```

This did not compile, because `[String]` did not conform to `JSValueCompatible`.

j-f1:
> It looks like fine to me 👍

Do you have a preference for it over the second proposal? I feel like the second one is more ergonomic.

kateinoigakukun:
@j-f1 
> This did not compile, because [String] did not conform to JSValueCompatible.
Ah, I got it. Makes sense to me to change the conformance condition.

> Do you have a preference for it over the second proposal? I feel like the second one is more ergonomic.
I prefer the second one.

j-f1:
Ok, I think I’m done making changes.

MaxDesiatov (approved):
Seems legit to me, thanks!

kateinoigakukun (approved):

MaxDesiatov:
>- [breaking] The `ConvertibleToJSValue` conformance on `Array` and `Dictionary` has been swapped from the `== ConvertibleToJSValue` case to the `: ConvertibleToJSValue` case.
>  - This means that e.g. `[String]` is now `ConvertibleToJSValue`, but `[ConvertibleToJSValue]` no longer conforms
>  - the `jsValue()` method still works in both cases
>  - to adapt existing code, use one of these approaches:
>    - use generics where possible (for single-type arrays)
>    - call `.map { $0.jsValue() }` (or `mapValues`) to get an array/dictionary of `JSValue` which you can then use as `ConvertibleToJSValue`
>    - add `.jsValue` to the end of all of the values in the array/dictionary literal

MaxDesiatov:
For reference, I'm trying to resolve build issues in https://github.com/TokamakUI/Tokamak/pull/471

j-f1:
It's definitely up for reconsideration. One option would be I guess to make special property wrappers for arrays/dictionaries of JSValueCompatible things on the DOMKit side. Otherwise there isn’t an alternative that I can think of.

PR #179, opened 2022-04-01
Add support for Symbol objects via JSSymbol

j-f1:
This PR adds support for the `Symbol` type, including both properties on the `Symbol` object itself and support for accessing Symbol-keyed properties on `JSObject` instances.

### DIFF
--- a/Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
+++ b/Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
@@ -83,7 +83,7 @@ public class JSFunction: JSObject {
     }
 }
 
-internal func invokeNonThrowingJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJSValue], this: JSObject?) -> RawJSValue {
+func invokeNonThrowingJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJSValue], this: JSObject?) -> RawJSValue {
     arguments.withRawJSValues { rawValues in
         rawValues.withUnsafeBufferPointer { bufferPointer in
             let argv = bufferPointer.baseAddress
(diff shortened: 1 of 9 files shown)


Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -84,18 +83,18 @@ public class JSFunction: JSObject {
     }
 }
 
-private func invokeNonThrowingJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJSValue], this: JSObject?) -> JSValue {
+internal func invokeNonThrowingJSFunction(_ jsFunc: JSFunction, arguments: [ConvertibleToJSValue], this: JSObject?) -> RawJSValue {
```
MaxDesiatov, line 86: In my understanding, `internal` is usually omitted as it's implicit. Is its presence warranted here?

Sources/JavaScriptKit/FundamentalObjects/JSSymbol.swift
```diff
@@ -0,0 +1,56 @@
...
+    public static let iterator: JSSymbol! = Symbol.iterator.symbol
+    public static let match: JSSymbol! = Symbol.match.symbol
+    public static let matchAll: JSSymbol! = Symbol.matchAll.symbol
+    public static let replace: JSSymbol! = Symbol.replace.symbol
+    public static let search: JSSymbol! = Symbol.search.symbol
+    public static let species: JSSymbol! = Symbol.species.symbol
+    public static let split: JSSymbol! = Symbol.split.symbol
+    public static let toPrimitive: JSSymbol! = Symbol.toPrimitive.symbol
+    public static let toStringTag: JSSymbol! = Symbol.toStringTag.symbol
+    public static let unscopables: JSSymbol! = Symbol.unscopables.symbol
```
MaxDesiatov, line 55: Wouldn't these be autogenerated from IDL in DOMKit?

j-f1: Unfortunately not, as the ECMAScript spec (where these are defined) does not use IDL. (And, technically, it isn’t part of the DOM — it’s a language feature available in Node/Deno/plain JavaScriptCore/etc)

MaxDesiatov (approved):
Seems legit to me 👍

PR #180, opened 2022-04-06
Reenable integration tests

kateinoigakukun:
Export `main` and execute `_initialize` and `main` in test entry

IntegrationTests are unintentionally disabled since
0a38d705e0c310f236599e2a0d62afa8012578db because reactor mode doesn't
export `_start` and `WASI.start` doesn't do anything except setting the
instance to WASI module. So export `main` manually, and call
`_initialize` according to the "new-style commands" described in
https://github.com/WebAssembly/WASI/blob/59cbe140561db52fc505555e859de884e0ee7f00/legacy/application-abi.md#current-unstable-abi

Symbol support (081784b34ad94c91eddccca58f1592ce181328ad) broke tests silently, so revert it for now.


kateinoigakukun:
CC: @j-f1

IntegrationTests/lib.js
```diff
@@ -45,6 +45,8 @@ const startWasiTask = async (wasmPath) => {
     swift.setInstance(instance);
     // Start the WebAssembly WASI instance!
     wasi.start(instance);
+    instance.exports._initialize();
+    instance.exports.main();
```
j-f1, line 49: Why isn’t this called by e.g. `wasi.start`? Is there a way to make sure existing users will make this change when upgrading?

kateinoigakukun: That's because @wasmer/wasi doesn't support reactor. But even though WASI implementation supports it, users need to call `WASI.initialize` instead: https://nodejs.org/api/wasi.html#wasiinitializeinstance.

> These kinds are mutually exclusive;

https://github.com/WebAssembly/WASI/blob/59cbe140561db52fc505555e859de884e0ee7f00/legacy/application-abi.md#current-unstable-abi

And they (command and reactor) are exclusive, so there is no way to upgrade without changing entrypoints

j-f1 (approved):

PR #181, opened 2022-04-08
Bump version to 0.14.0, update `CHANGELOG.md`

MaxDesiatov:
New release is required to resolve https://github.com/TokamakUI/Tokamak/pull/475.

I've also cleaned up formatting in `.ts` file and clarified version incompatibility error message.

### DIFF
--- a/.swift-version
+++ b/.swift-version
@@ -1,1 +1,1 @@
-wasm-5.5.0-RELEASE
+wasm-5.6.0-RELEASE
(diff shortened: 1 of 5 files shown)


kateinoigakukun (approved):

kateinoigakukun:
I think this new release is not required for the Tokamak PR. Is there any essential change?

MaxDesiatov:
https://github.com/swiftwasm/carton/pull/310 relies on reactor model, which was updated in https://github.com/swiftwasm/JavaScriptKit/pull/180, wasn't it? And then `.jsValue()` -> `.jsValue` change was breaking, I'd like to update Tokamak codebase for it sooner rather than later.

kateinoigakukun:
#180 is only for test harnesses, so no change in the js runtime. Yeah, it's fine to release breaking changes soon.

PR #183, opened 2022-04-08
Re-add support for Symbol objects via JSSymbol

j-f1:
This re-adds the functionality removed by #180.

(I accidentally pushed this change to `main` first, sorry about that!)

The [`Reflect` namespace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) is mainly intended for use with `Proxy` objects, with each function on that namespace matching the signature of the corresponding proxy trap. For most use cases, normal access is preferable.

The test failure was occurring because `Reflect.*` functions check that their parameter is really an object. While this is the case for objects, symbols aren’t objects; the `[]` syntax automatically converts to an object where necessary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 /*.xcodeproj
 xcuserdata/
 .swiftpm
+.vscode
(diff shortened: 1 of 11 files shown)


MaxDesiatov (commented):
LGTM 👍

MaxDesiatov (approved):

kateinoigakukun:
For the record: https://tc39.es/ecma262/#sec-reflect.get
> 28.1.5 Reflect.get ( target, propertyKey [ , receiver ] )
> When the get function is called with arguments target, propertyKey, and receiver, the following steps are taken:
> 1. If [Type](https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values)(target) is not Object, throw a TypeError exception.

kateinoigakukun (approved):
Thanks :)

Issue #116, opened 2021-01-06
Add an implementation for `JSValueEncoder`


MaxDesiatov:
Do I understand correctly that there's no easy way right now to convert an arbitrary `Encodable` type to `JSValue`? We already have `JSValueDecoder`, but apparently lack the opposite...

jhoughjr:
What all would this entail? I assume an inverse of the decoding method?

MaxDesiatov:
Yes, I would assume so, but it may need some investigation to confirm this.

jhoughjr:
Ive discovered a bug it seems that might relate to this issue.

working with Request and Fetch apis.
long story short no matter what body I send to my server, JavaScriptKit seems to ensure it is the string literal [object Object] even if I give it my own string literal.
Thats not a log output either, its the decode utf8 data on the request at the server.

I tested this by setting body to a JSValue(stringLiteral: ) of the exact json body I wanted.
What does my served decode the 15 bytes as [object Object]
This is the exact same output as body = body.description
info is available for the last few days in the tokamak channel.

So if it cant send the body I tell it to, Encoding won't matter cuz it will just encode the literal string "[object Object]"

ephemer:
@kateinoigakukun @j-f1 FYI I have created an implemention of JSValueEncoder. I can make a PR soon – just testing it some more first with some more use-cases.

PR #185, opened 2022-04-09
Fix JSValueDecoder

j-f1:
We didn’t have any tests exercising this code, and as far as I can tell it has never worked.


MaxDesiatov (approved):
Nice catch! Didn't we have an open issue for this somewhere, or am I misremembering?

j-f1:
Not that I know of, although there is #116 for implementing `JSValueEncoder`.

PR #186, opened 2022-04-09
Cleanup & improvements to perf-tester

j-f1:
- clean up the code to remove unnecessary configuration options and other cruft from the original action
- quadruple the number of total benchmark runs, hopefully reducing the amount of noise regularly seen in benchmark runs
- remove false precision in the reported numbers
- tweak the automatic comment slightly


MaxDesiatov (approved):
LGTM 👍

PR #190, opened 2022-05-04
Gracefully handle unavailable `JSBridgedClass`

MaxDesiatov:
Force unwrapping in `class var constructor` may crash for classes that are unavailable in the current environment. For example, after https://github.com/swiftwasm/WebAPIKit/pull/38 was merged, it led to crashes in `getCanvas` calls due to these optional casts relying on `class var constructor` always succeeding:

```swift
    public static func construct(from value: JSValue) -> Self? {
        if let canvasRenderingContext2D: CanvasRenderingContext2D = value.fromJSValue() {
            return .canvasRenderingContext2D(canvasRenderingContext2D)
        }
        if let gpuCanvasContext: GPUCanvasContext = value.fromJSValue() {
            return .gpuCanvasContext(gpuCanvasContext)
        }
        if let imageBitmapRenderingContext: ImageBitmapRenderingContext = value.fromJSValue() {
            return .imageBitmapRenderingContext(imageBitmapRenderingContext)
        }
        if let webGL2RenderingContext: WebGL2RenderingContext = value.fromJSValue() {
            return .webGL2RenderingContext(webGL2RenderingContext)
        }
        if let webGLRenderingContext: WebGLRenderingContext = value.fromJSValue() {
            return .webGLRenderingContext(webGLRenderingContext)
        }
        return nil
    }
```

`if let gpuCanvasContext: GPUCanvasContext = value.fromJSValue()` branch crashes on browsers that don't have `GPUCanvasContext` enabled.

As we currently don't have a better way to handle unavailable features, I propose making the result type of `static var constructor` requirement optional. This means you can still declare classes that are unavailable in the host JS environment. Conditional type casts are also available as they were, they will just always return `nil`, and initializers for these classes will return `nil` as well.

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSArray.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSArray.swift
@@ -1,10 +1,10 @@
 /// A wrapper around [the JavaScript Array class](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array)
 /// that exposes its properties in a type-safe and Swifty way.
 public class JSArray: JSBridgedClass {
-    public static let constructor = JSObject.global.Array.function!
+    public static let constructor = JSObject.global.Array.function
 
     static func isArray(_ object: JSObject) -> Bool {
-        constructor.isArray!(object).boolean!
+        constructor!.isArray!(object).boolean!
     }
 
     public let jsObject: JSObject
@@ -94,8 +94,8 @@ private func getObjectValuesLength(_ object: JSObject) -> Int {
     return Int(values.length.number!)
 }
 
-extension JSValue {
-    public var array: JSArray? {
+public extension JSValue {
+    var array: JSArray? {
         object.flatMap(JSArray.init)
     }
 }
--- a/Sources/JavaScriptKit/BasicObjects/JSDate.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSDate.swift
@@ -1,32 +1,32 @@
-/** A wrapper around the [JavaScript Date 
-class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) that
-exposes its properties in a type-safe way. This doesn't 100% match the JS API, for example 
-`getMonth`/`setMonth` etc accessor methods are converted to properties, but the rest of it matches
-in the naming. Parts of the JavaScript `Date` API that are not consistent across browsers and JS 
-implementations are not exposed in a type-safe manner, you should access the underlying `jsObject`
-property if you need those.
-*/
+/** A wrapper around the [JavaScript Date
+ class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) that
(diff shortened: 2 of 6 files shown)


Sources/JavaScriptKit/BasicObjects/JSDate.swift
```diff
@@ -1,32 +1,32 @@
...
-property if you need those.
-*/
+/** A wrapper around the [JavaScript Date
+ class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) that
+ exposes its properties in a type-safe way. This doesn't 100% match the JS API, for example
+ `getMonth`/`setMonth` etc accessor methods are converted to properties, but the rest of it matches
+ in the naming. Parts of the JavaScript `Date` API that are not consistent across browsers and JS
+ implementations are not exposed in a type-safe manner, you should access the underlying `jsObject`
+ property if you need those.
+ */
```
MaxDesiatov, line 8: Fix applied by the formatter.

Sources/JavaScriptKit/BasicObjects/JSArray.swift
```diff
@@ -94,8 +94,8 @@ private func getObjectValuesLength(_ object: JSObject) -> Int {
     return Int(values.length.number!)
 }
 
-extension JSValue {
-    public var array: JSArray? {
+public extension JSValue {
```
MaxDesiatov, line 97: Fix applied by the formatter.

MaxDesiatov:
Trying this in https://github.com/swiftwasm/WebAPIKit/pull/41, looks like there's more work to do, but I'm still looking forward to any feedback on this as a general approach.

Sources/JavaScriptKit/BasicObjects/JSArray.swift
```diff
@@ -1,10 +1,10 @@
 /// A wrapper around [the JavaScript Array class](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Array)
 /// that exposes its properties in a type-safe and Swifty way.
 public class JSArray: JSBridgedClass {
-    public static let constructor = JSObject.global.Array.function!
+    public static let constructor = JSObject.global.Array.function
```
j-f1, line 4: For types like these where the global must be available, how do you feel about declaring `constructor` as `JSFunction!`?

j-f1:
Overall I like this! I had been contemplating adding an Boolean `supported` class var but this is a much better approach.

Sources/JavaScriptKit/BasicObjects/JSArray.swift
MaxDesiatov: I tried this, but then you have to add an explicit `JSFunction?` type signature on top, which means you still have to unwrap with `!` or `?` wherever this `constructor` is used. So after all, there's no benefit of unwrapping it and then wrapping back again into `JSFunction?` to satisfy the protocol requirement.

MaxDesiatov:
I've verified that this fully works with WebAPIKit, ready for review.

j-f1 (approved):

kateinoigakukun (approved):
Ah, I missed this PR

PR #193, opened 2022-05-12
Supply JSKit runtime in SwiftPM resources

MaxDesiatov:
Due to the fact that https://github.com/apple/swift-package-manager/pull/4306 will only be available starting with Swift 5.7, we can't generate these resources with a SwiftPM plugin. I propose storing the generated code in the repository directly for now. When 5.7 is released, we can switch to a SwiftPM plugin approach.

Can be tested end-to-end with https://github.com/swiftwasm/carton/pull/335.


kateinoigakukun (commented):
Could you add [`linguist-generated` entry in `.gitattributes` to ignore the generated file from stats](https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes).
Also, could we have a script to update `index.{js,mjs}` and check that they are up-to-date in CI?

j-f1 (approved):

.github/workflows/test.yml
```diff
@@ -38,6 +38,10 @@ jobs:
           SWIFT_VERSION=${{ matrix.entry.toolchain }} make bootstrap
           echo ${{ matrix.entry.toolchain }} > .swift-version
           make test
+      - name: Check if SwiftPM resources are stale
+        run:
+          git diff --exit-code Sources/JavaScriptKit/Runtime
```
kateinoigakukun, line 43:
```suggestion
          make regenerate_swiftpm_resources
          git diff --exit-code Sources/JavaScriptKit/Runtime
```

Missing the line?

.gitattributes
```diff
@@ -0,0 +1 @@
+Sources/JavaScriptKit/Runtime linguist-generated
```
kateinoigakukun, line 1:
```suggestion
Sources/JavaScriptKit/Runtime/* linguist-generated
```

According to https://git-scm.com/docs/gitattributes
> patterns that match a directory do not recursively match paths inside that directory (so using the trailing-slash path/ syntax is pointless in an attributes file; use path/** instead)

kateinoigakukun (approved):

PR #195, opened 2022-05-16
Improve JSKit diagnostics for use-after-free of JSClosure

kateinoigakukun:
The current JSKit’s diagnostics for use-after-free of JSClosure is unclear for developers to find which JSClosure is used. 
To mitigate the debugging difficulties, JSKit should provide more information in diagnostics for use-after-free of JSClosure

Before
> The function was already released
> Unreachable code should not be executed

After
> Error: The JSClosure was already released by Swift side. The closure is created at PrimaryTests/main.swift:247


MaxDesiatov (approved):
👍

PR #200, opened 2022-07-19
Add async JSPromise.result property

kateinoigakukun:
Add `JSPromise.result`, which is similar with `Task.result`

### DIFF
--- a/Sources/JavaScriptEventLoop/JavaScriptEventLoop.swift
+++ b/Sources/JavaScriptEventLoop/JavaScriptEventLoop.swift
@@ -116,6 +116,24 @@ public extension JSPromise {
             }
         }
     }
+
+    /// Wait for the promise to complete, returning its result or exception as a Result.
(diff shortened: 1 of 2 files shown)


Sources/JavaScriptEventLoop/JavaScriptEventLoop.swift
```diff
@@ -116,6 +116,24 @@ public extension JSPromise {
             }
         }
     }
+
+    /// Wait for the promise to complete, returning its result or exception as a Result.
+    var result: Result<JSValue, JSValue> {
```
yonihemi, line 121: Nice!
Don't we want this to be public?

MaxDesiatov: The extension is public, which makes everything declared within it public.

yonihemi (approved):

PR #203, opened 2022-08-12
Refine benchmark suite

kateinoigakukun:
We had been running benchmarks for a long time and they were very unstable. And also couldn't find meaningful results.

I changed to perform benchmarks with release configuration, and don't repeat noisy prelude for each benchmark.
Also added several benchmark cases to find our bridging bottlenecks (assignment, call, value conversion, ...)


MaxDesiatov (approved):
These are great improvements, thank you!

PR #204, opened 2022-08-12
Add Int64/UInt64 to Bigint slow conversion

kateinoigakukun:
Int64 -> BigInt conversion can be done by passing two i32 without i64-bigint integration feature. This is a slow path, but useful for environments that have BigInt but don't have i64-bigint integration.


MaxDesiatov (approved):
Overall LGTM. Do I understand correctly that the choice between the slow path and fast path is made explicitly by the user here?

kateinoigakukun:
Yes, the variant choice would be made in caller side for now.

kateinoigakukun:
BTW, some of the benchmark cases are too fast and cause meaningless reports...

PR #207, opened 2022-08-16
Runtime Performance Optimization

kateinoigakukun:
- Reduce unnecessary calls of `memory` and `dataView` getter in runtime TS code
- Promote result pointer to return value to reduce memory operations
- Reduce Swift ARC retain/release operations by:
  - Reduce Optional wrapper
  - Reduce reference escapes
- Add fast paths for empty arguments call

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -84,6 +84,7 @@ export class SwiftRuntime {
     ) {
         const argc = args.length;
         const argv = this.exports.swjs_prepare_host_function_call(argc);
+        const memory = this.memory;
         for (let index = 0; index < args.length; index++) {
             const argument = args[index];
             const base = argv + 16 * index;
@@ -93,12 +94,12 @@ export class SwiftRuntime {
                 base + 4,
                 base + 8,
                 false,
-                this.memory
+                memory
             );
         }
         let output: any;
         // This ref is released by the swjs_call_host_function implementation
-        const callback_func_ref = this.memory.retain((result: any) => {
+        const callback_func_ref = memory.retain((result: any) => {
             output = result;
         });
         const alreadyReleased = this.exports.swjs_call_host_function(
@@ -127,9 +128,10 @@ export class SwiftRuntime {
             payload1: number,
             payload2: number
         ) => {
-            const obj = this.memory.getObject(ref);
-            const key = this.memory.getObject(name);
-            const value = JSValue.decode(kind, payload1, payload2, this.memory);
+            const memory = this.memory;
+            const obj = memory.getObject(ref);
+            const key = memory.getObject(name);
+            const value = JSValue.decode(kind, payload1, payload2, memory);
             obj[key] = value;
         },
         swjs_get_prop: (
@@ -139,8 +141,9 @@ export class SwiftRuntime {
             payload1_ptr: pointer,
             payload2_ptr: pointer
         ) => {
-            const obj = this.memory.getObject(ref);
-            const key = this.memory.getObject(name);
+            const memory = this.memory;
+            const obj = memory.getObject(ref);
+            const key = memory.getObject(name);
             const result = obj[key];
             JSValue.write(
                 result,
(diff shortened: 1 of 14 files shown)


Runtime/src/index.ts
```diff
@@ -309,37 +284,21 @@ export class SwiftRuntime {
...
-            kind_ptr: pointer,
             payload1_ptr: pointer,
             payload2_ptr: pointer
         ) => {
             const obj = this.memory.getObject(obj_ref);
             const func = this.memory.getObject(func_ref);
-            let isException = true;
-            try {
+            let result = undefined;
                 const args = JSValue.decodeArray(argv, argc, this.memory);
```
sunnypp, line 293: minor: indentation

Runtime/src/index.ts
```diff
@@ -127,28 +121,28 @@ export class SwiftRuntime {
...
             payload2_ptr: pointer
         ) => {
-            const obj = this.memory.getObject(ref);
-            const key = this.memory.getObject(name);
+            const memory = this.memory;
+            const obj = memory.getObject(ref);
+            const key = memory.getObject(name);
             const result = obj[key];
-            JSValue.write(
+            return JSValue.writeV2(
```
MaxDesiatov, line 140: Is the old version of `write` still used anywhere? Should it be renamed to `writeV1` then?

kateinoigakukun: The old one is still used for writing out multiple values (e.g. passing js values to Swift closure )

Runtime/src/index.ts
```diff
@@ -127,28 +121,28 @@ export class SwiftRuntime {
...
-            const obj = this.memory.getObject(ref);
-            const key = this.memory.getObject(name);
-            const value = JSValue.decode(kind, payload1, payload2, this.memory);
+            const memory = this.memory;
+            const obj = memory.getObject(ref);
+            const key = memory.getObject(name);
+            const value = JSValue.decode(kind, payload1, payload2, memory);
             obj[key] = value;
         },
         swjs_get_prop: (
```
MaxDesiatov, line 130: Would you mind adding doc comments to functions in `wasmImports`, including their parameters and return values/types? Thanks!

Sources/_CJavaScriptKit/include/_CJavaScriptKit.h
```diff
@@ -175,66 +178,72 @@ extern JavaScriptObjectRef _i64_to_bigint_slow(unsigned int lower, unsigned int
...
 
 /// `_call_function` calls JavaScript function with given arguments list without capturing any exception
 ///
 /// @param ref The target JavaScript function to call.
 /// @param argv A list of `RawJSValue` arguments to apply.
 /// @param argc The length of `argv``.
-/// @param result_kind A result pointer of JavaScript value kind of returned result or thrown exception.
 /// @param result_payload1 A result pointer of first payload of JavaScript value of returned result or thrown exception.
 /// @param result_payload2 A result pointer of second payload of JavaScript value of returned result or thrown exception.
+/// @return A JavaScriptValueKindAndFlags bits represented as 32bit integer for the returned value.
```
MaxDesiatov, line 200:
```suggestion
/// @return A `JavaScriptValueKindAndFlags` bits represented as 32bit integer for the returned value.
```

MaxDesiatov (approved):
Fantastic work, these are very impressive improvements! 👏

I only have a couple of suggestions for improving doc comments, but otherwise this LGTM.

Runtime/src/index.ts
kateinoigakukun: They would be exactly same as `_CJavaScriptKit.h` one. Do you mean we need to add docs here as well for library users?

MaxDesiatov: Ah, makes sense. Then we probably don't need those duplicated I think. Thanks for checking!

MaxDesiatov (approved):

PR #215, opened 2022-10-08
Use swiftwasm/setup-swiftwasm instead of swiftenv on CI

kateinoigakukun:


kateinoigakukun:
The old main branch expects to have swiftenv, so it's failing now. But it will be back to green after merging this.

Makefile
```diff
@@ -2,7 +2,6 @@ MAKEFILE_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
 
 .PHONY: bootstrap
 bootstrap:
-	./scripts/install-toolchain.sh
```
j-f1, line 5: Can this script file be deleted too?

j-f1 (approved):

Makefile
kateinoigakukun: It's still useful for installing the default toolchain without carton, and I still use it.

Issue #218, opened 2022-12-19
JSClosure will be overriden from other closure


sidepelican:
JSClosure stores its identifier and tell it to JS for bridging, but the identifiers has a trouble.
`hashValue` is used for the identifier, but `hashValue` is not guaranteed to be unique.

https://github.com/swiftwasm/JavaScriptKit/blob/1afbfaf16598119a851dca39e3b447de6cf52589/Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift#L89

j-f1:
Hm, good point! Looking at [the relevant implementations](https://github.com/apple/swift/blob/main/stdlib/public/core/ObjectIdentifier.swift) I think that should be replaced with the `UInt(bitPattern:)` initializer which can take an `ObjectIdentifier`.

PR #219, opened 2022-12-19
Prefer `UInt(bitPattern:)` for object id to guarantee uniqueness

kateinoigakukun:
Close https://github.com/swiftwasm/JavaScriptKit/issues/218


sidepelican:
Thanks!

PR #232, opened 2024-02-06
Fix availability marker for Swift 5.9 compiler targeting host machine

kateinoigakukun:
When building JavaScriptKit with Xcode, targeting host machine, the `func enqueue(_: ExecutorJob)` must be guarded by `@available` attribute explicitly. And due to the 5.9 compiler issue, it emit some migration warnings too consevatively. This commit suppress the warning by updating minimum available OS versions. Those versions should not be a problem when building for Wasm.


fjtrujy (approved):
Tested myself in my project, it is working

PR #236, opened 2024-03-24
[CI] Update actions and configure Dependabot

ikesyo:


.github/workflows/compatibility.yml
```diff
@@ -9,9 +9,7 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 1
+        uses: actions/checkout@v4
```
ikesyo, line 12: `fetch-depth: 1` by default: https://github.com/actions/checkout/blob/cd7d8d697e10461458bc61a30d094dc601a8b017/action.yml#L70-L72

kateinoigakukun (approved):

PR #237, opened 2024-03-25
Replace `@_spi` usages with `package` access modifier

ikesyo:
- [SE-0386](https://github.com/apple/swift-evolution/blob/main/proposals/0386-package-access-modifier.md)


kateinoigakukun:
Actually, there are some use cases of this SPI in some real applications to bypass some overhead (e.g. argument and method name serializations) while sharing object id space. So I'd like to keep this exposed as a "feature" for expert.

ikesyo:
Thanks for the detailed explanation! 👍

PR #240, opened 2024-04-07
Fix `JSClosure` leak

kateinoigakukun:
```swift
let c1 = JSClosure { _ in .undefined }
consume c1
```

did not release the `JSClosure` itself and it leaked the underlying
JavaScript closure too because `JSClosure` -> JS closure thunk ->
Closure registry entry -> `JSClosure` reference cycle was not broken
when using FinalizationRegistry. (Without FR, it was broken by manual
`release` call.)

Note that weakening the reference does not violates the contract that
function reference should be unique because holding a weak reference does
deinit but not deallocate the object, so ObjectIdentifier is not reused
until the weak reference in the registry is removed.


kateinoigakukun:
Tested with a random large-scale application and no major regression found

PR #256, opened 2024-07-06
Add `WebWorkerTaskExecutor`

kateinoigakukun:
WebWorkerTaskExecutor is an implementation of `TaskExecutor` protocol, which is introduced by [SE-0417] since Swift 6.0. This task executor runs tasks on Worker threads, which is useful for offloading computationally expensive tasks from the main thread.

The `WebWorkerTaskExecutor` is designed to work with [Web Workers API] and Node.js's [`worker_threads` module].

This depends on https://github.com/swiftlang/swift/pull/75008

## Example

Try it out: https://swiftwasm-threading-example.vercel.app/

Source: https://github.com/kateinoigakukun/swiftwasm-threading-example/

## Usage

```swift
import JavaScriptEventLoop

JavaScriptEventLoop.installGlobalExecutor()
WebWorkerTaskExecutor.installGlobalExecutor()

func render() async {
  let executor = WebWorkerTaskExecutor(numberOfThreads: 8)
  defer { executor.terminate() }

  await withTaskGroup(of: Void.self) { group in
      let yStride = scene.height / concurrency
      for i in 0..<concurrency {
          let yRange = i * yStride..<(i + 1) * yStride
          let work = Work(scene: scene, imageView: imageView, yRange: yRange)
          group.addTask(executorPreference: executor) { work.run() }
      }
      if scene.height % concurrency != 0 {
          let work = Work(scene: scene, imageView: imageView, yRange: (concurrency * yStride)..<scene.height)
          group.addTask(executorPreference: executor) { work.run() }
      }
  }
}
```

Also JavaScript-side needs some tweaks:

```javascript
// --- main.js
class ThreadRegistry {
  workers = new Map();
  nextTid = 1;

  spawnThread(worker, module, memory, startArg) {
    const tid = this.nextTid++;
    this.workers.set(tid, worker);
    worker.postMessage({ module, memory, tid, startArg });
    return tid;
  }

  listenMainJobFromWorkerThread(tid, listener) {
    const worker = this.workers.get(tid);
    worker.onmessage = (event) => {
      listener(event.data);
    };
  }

  wakeUpWorkerThread(tid, data) {
    const worker = this.workers.get(tid);
    worker.postMessage(data);
  }
}

const threads = new ThreadRegistry();
const swift = new SwiftRuntime({
  threadChannel: {
    wakeUpWorkerThread: threads.wakeUpWorkerThread.bind(threads),
    listenMainJobFromWorkerThread: threads.listenMainJobFromWorkerThread.bind(threads)
  }
});

const module = await WebAssembly.compile("./main.wasm");
const memory = new WebAssembly.Memory({ initial: 256, maximum: 16384, shared: true })

const importObject = {
  wasi_snapshot_preview1: wasi.wasiImport,
  javascript_kit: swift.wasmImports,
  env: { memory },
  wasi: {
    "thread-spawn": (startArg) => {
      const worker = new Worker("./worker.js", { type: "module" });
      return threads.spawnThread(worker, module, memory, startArg);
    }
  }
}

// Instantiate the WebAssembly file
const instance = await WebAssembly.instantiate(module, importObject);

swift.setInstance(instance);
// Start the WebAssembly WASI instance!
wasi.start(instance, swift);


// --- worker.js 

self.onmessage = (event) => {
  self.onmessage = null;

  const swift = new SwiftRuntime({
    threadChannel: {
      wakeUpMainThread: (unownedJob) => {
        // Send the job to the main thread
        postMessage(unownedJob);
      },
      listenWakeEventFromMainThread: (listener) => {
        self.onmessage = (event) => listener(event.data);
      }
    }
  }


  const { module, memory, tid, startArg } = event.data;

  const importObject = {
    wasi_snapshot_preview1: wasi.wasiImport,
    javascript_kit: swift.wasmImports,
    env: { memory },
    wasi: {
      "thread-spawn": (startArg) => {
         throw new Error("Cannot spawn a new thread from a worker thread")
      }
    }
  }

  const instance = await WebAssembly.instantiate(module, importObject);
  swift.setInstance(instance);
  wasi.setInstance(instance);
  swift.startThread(tid, startArg);
}
```

## Dataflow overview

WebWorkerTaskExecutor drawio (1)

### `@MainActor` does not hop back to main thread

Due to an issue in [Cooperative global executor](https://github.com/swiftlang/swift/blob/release/6.0/stdlib/public/Concurrency/CooperativeGlobalExecutor.inc), `@MainActor` and `MainActor.run` don't switch execution thread when `WebWorkerTaskExecutor` is preferred.

### `JSObject` instance cannot cross the thread boundary

Due to the underlying Web Worker limitation, JavaScript objects (`JSObject`) cannot be shared with or transferred to another thread. You need to convert it into Swift-native objects to represent it in shared memory space.

```swift
let canvas = JSObject.global.document.getElementById("my-canvas")
Task(executorPreference: executor) {
  let context = canvas.getContext("2d") // Invalid!
}
```

## TODO

- Job scheduling should take into account Task priority
- Better scheduling algorithm
  - e.g. Schedule jobs to other workers when the current worker thread is busy (the current scheduler enqueues jobs to the current worker thread)
- Rewrite JSClosure not to use global storage
- Add a way to transfer or clone objects across workers
- Worker-backed SerialiExecutor implementation?
  - It would be useful to bind an actor to a dedicated worker. In such actor context, we can safely access JS objects

[SE-0417]: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0417-task-executor-preference.md
[Web Workers API]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API
[`worker_threads` module]: https://nodejs.org/api/worker_threads.html

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,16 +23,16 @@ jobs:
           - { os: ubuntu-20.04, toolchain: wasm-5.9.1-RELEASE, wasi-backend: MicroWASI }
           - { os: ubuntu-20.04, toolchain: wasm-5.10.0-RELEASE, wasi-backend: MicroWASI }
           - os: ubuntu-22.04
-            toolchain: DEVELOPMENT-SNAPSHOT-2024-05-01-a
+            toolchain: DEVELOPMENT-SNAPSHOT-2024-06-13-a
             swift-sdk:
-              id: DEVELOPMENT-SNAPSHOT-2024-05-25-a-wasm32-unknown-wasi
-              download-url: "https://github.com/swiftwasm/swift/releases/download/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-05-25-a/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-05-25-a-wasm32-unknown-wasi.artifactbundle.zip"
+              id: DEVELOPMENT-SNAPSHOT-2024-06-14-a-wasm32-unknown-wasi
+              download-url: "https://github.com/swiftwasm/swift/releases/download/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-06-14-a/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-06-14-a-wasm32-unknown-wasi.artifactbundle.zip"
             wasi-backend: Node
           - os: ubuntu-22.04
-            toolchain: DEVELOPMENT-SNAPSHOT-2024-05-01-a
+            toolchain: DEVELOPMENT-SNAPSHOT-2024-06-13-a
             swift-sdk:
-              id: DEVELOPMENT-SNAPSHOT-2024-05-25-a-wasm32-unknown-wasip1-threads
-              download-url: "https://github.com/swiftwasm/swift/releases/download/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-05-25-a/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-05-25-a-wasm32-unknown-wasip1-threads.artifactbundle.zip"
+              id: DEVELOPMENT-SNAPSHOT-2024-06-14-a-wasm32-unknown-wasip1-threads
+              download-url: "https://github.com/swiftwasm/swift/releases/download/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-06-14-a/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-06-14-a-wasm32-unknown-wasip1-threads.artifactbundle.zip"
             wasi-backend: Node
 
     runs-on: ${{ matrix.entry.os }}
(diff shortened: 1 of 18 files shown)


kateinoigakukun:
CC: @ephemer might be interested in :)

ephemer:
Hi @kateinoigakukun thanks, just saw this in the JSKit code and randomly saw your ping here after investigating more closely. I don't get notifications from GitHub, sorry! 

This is interesting, and definitely makes sense to me as an approach for Swift concurrency. Our use case is to put an ML workload onto a background thread. That workload is managed by tfjs, and we can't do much more with splitting up the work into chunks like in the example here. I think we'd have to use "real" threading for our use case to benefit from this direction of work. I think generally that would be the best outcome for Swift Wasm too (using "real" Wasm threads). I know that not all runtime environments support it though (including ours – we need to support browsers back to ~Safari 12)

PR #263, opened 2024-10-08
WIP: Support for Embedded Swift (v2)

sliemeobn:
thanks to @MihaelIsaev for starting the other branch, I was inspired to try on my own by it.

I left String untouched (the entire JS runtime really), but I had to move a few things out of the C lib (ideally we can get rid of the entire thing eventually).

Also, the C header bit field with stuff does not seem to play well with Embedded (do not know why, pushed it into Swift for now)

there is an `Examples/Embedded` demo that shows how it wants to be built.

my hope is that, one day:
1) SwiftPM learns to set the flags on its own with a dedicated "Embedded" mode
2) variadic generics will be supported

for now, a single unsafe flag needs to be added in the package.swift, and bundle resources do not work (SwiftPM will generate code for Foundation bundles).

In general, JavaScriptKit could use a bit of modernization and replace of all the "anys" with variadic generics.
Also, I don't see the value of using class hierarchies (does not play well with embedded) - but I tried to leave as much untouched as possible.

Promises, throwing functions, and typed arrays are left unsupported entirely for now, the rest should work with almost no API surface change.


sliemeobn:
also @MihaelIsaev, I think your Foundation package is a great idea and a great start. 
for whatever reason I ended up needing `strlen` as well, and `memmove` is needed to get string interpolation going.

do you think you could add these from the wasi/musl libc as well, and add proper license attribution to your package?
also, IMHO the name "Foundation" is very confusing. maybe we can come up with a better name?

Sources/JavaScriptKit/FundamentalObjects/JSObject.swift
```diff
@@ -224,3 +229,32 @@ public class JSThrowingObject {
         self[name]
     }
 }
+#endif
+
+public protocol _JSObjectProtocol: JSObject {
```
kateinoigakukun, line 234: Could you elaborate on why this protocol is needed in its doc comment?

Sources/JavaScriptKit/FundamentalObjects/JSSymbol.swift
```diff
@@ -10,7 +10,11 @@ public class JSSymbol: JSObject {
 
     public init(_ description: JSString) {
         // can’t do `self =` so we have to get the ID manually
+    #if hasFeature(Embedded)
```
kateinoigakukun, line 13: Can we unconditionally use `description.jsValue` here?

Examples/Embedded/Sources/EmbeddedApp/_thingsThatShouldNotBeNeeded.swift
```diff
@@ -0,0 +1,29 @@
+import JavaScriptKit
+
+// NOTE: it seems the embedded tree shaker gets rid of these exports if they are not used somewhere
+func _i_need_to_be_here_for_wasm_exports_to_work() {
```
kateinoigakukun, line 4: Could you try adding `@_used` (for Swift decls) or `__attribute__((used))` (for C decls) to teach LTO not to DCE those decls?

Sources/JavaScriptKit/FundamentalObjects/JSClosure.swift
```diff
@@ -234,3 +235,20 @@ func _free_host_function_impl(_ hostFuncRef: JavaScriptHostFuncRef) {
     JSClosure.sharedClosures[hostFuncRef] = nil
 }
 #endif
+
+#if hasFeature(Embedded)
+// cdecls currently don't work in embedded, and expose for wasm only works >=6.0
```
kateinoigakukun, line 240: There is no reason using `cdecl` instead of `silgen_name` here, so I think we can use `silgen_name` and call those functions with Swift calling convention. I'll address this after merging your change, so don't care about this comment, just a note for me.

kateinoigakukun:
I think it's a much mergeable approach. Happy to merge once CI passed and a few my concerns are resolved. Thank you for tackling on this!

sliemeobn:
thanks for reviewing, I'll try address the comments later today, they all makes sense.

the biggest issue with 5.10 compatibility I have right now is that the swift compiler seems to freak out about `@_expose(wasm)` even when it is `#if`ed away.

I'll try my luck with `@_used` on the cdecls later, maybe we stick to the C shims for exposing to wasm untouched for now...

Examples/Embedded/Sources/EmbeddedApp/_thingsThatShouldNotBeNeeded.swift
sliemeobn: I tried `@_used` for both `@_expose` and `@_cdecl`, does not work : (
I'll post in the forums about it, maybe this can be fixed in the future somehow.
An underscored public is is the best I can come up with atm.

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -79,11 +85,21 @@ public class JSFunction: JSObject {
         JSThrowingFunction(self)
     }
 
-    /// A variadic arguments version of `new`.
-    public func new(_ arguments: ConvertibleToJSValue...) -> JSObject {
-        new(arguments: arguments)
+#else
+    @discardableResult
+    public func callAsFunction(arguments: [JSValue]) -> JSValue {
```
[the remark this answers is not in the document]
sliemeobn: I shied away from it not to add potentially conflicting overloads.
But this should be fine to add.

Sources/JavaScriptKit/FundamentalObjects/JSSymbol.swift
sliemeobn: we sure can

Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
```diff
@@ -93,43 +109,155 @@ public class JSFunction: JSObject {
...
+    }
+
+    func new(_ arg0: some ConvertibleToJSValue, _ arg1: some ConvertibleToJSValue, _ arg2: some ConvertibleToJSValue, _ arg3: some ConvertibleToJSValue, _ arg4: some ConvertibleToJSValue, _ arg5: some ConvertibleToJSValue, _ arg6: some ConvertibleToJSValue) -> JSObject {
+        new(arguments: [arg0.jsValue, arg1.jsValue, arg2.jsValue, arg3.jsValue, arg4.jsValue, arg5.jsValue, arg6.jsValue])
+    }
+}
+
+// C bit fields seem to not work with Embedded
+// in "normal mode" this is defined as a C struct
+private struct JavaScriptValueKindAndFlags {
```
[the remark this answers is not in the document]
sliemeobn: just to clarify, do you mean to use the swift-based "decoding" as I wrote here for embedded in general? (ie: just pass an "unsigned short" through the C shim?)

sliemeobn:
@kateinoigakukun with CI passing, are there any issues left you'd want me to address before considering a merge?

kateinoigakukun:
Sorry I just missed the CI status update

kateinoigakukun (approved):

PR #269, opened 2024-11-17
Added discardableresult 

Amzd:
This prevents the warnings for unused results through `@discardableResult` which means you no longer need `_ =` everywhere.

TODO: Update comments and readme


kateinoigakukun (approved):
Thanks!

PR #270, opened 2024-11-17
nonmutating

### LINKED ISSUE #132: Cannot set HTMLElement.style properties

Amzd:
Fixes #132

### DIFF
--- a/Sources/JavaScriptKit/JSValue.swift
+++ b/Sources/JavaScriptKit/JSValue.swift
@@ -116,6 +116,7 @@ public extension JSValue {
         nonmutating set { self.object![name] = newValue }
     }
 
+#if !hasFeature(Embedded)
     /// An unsafe convenience method of `JSObject.subscript(_ index: String) -> JSValue`
     /// - Precondition: `self` must be a JavaScript Object.
     @_disfavoredOverload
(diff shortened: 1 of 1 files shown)


Amzd:
Hmm okay I guess it is not as easy as that (tests do not complete), but it works for my use case. I don't know how to debug this issue as the tests crash without this PR on my local machine too.

kateinoigakukun:
Closing in favor of https://github.com/swiftwasm/JavaScriptKit/pull/321

PR #278, opened 2024-12-08
Fix empty TypedArray creation

kateinoigakukun:
Close https://github.com/swiftwasm/swift/issues/5599

Empty array instances share the same storage for all element types, so their `baseAddress` can be unaligned. (It seems the empty array storage had been placed at 8-aligned position fortunately but it's no longer the case since Swift 6.0, I guess)

https://github.com/swiftlang/swift/blob/2914d0e46eda33cadfd81c3381e8737855ce397c/stdlib/public/stubs/GlobalObjects.cpp#L42


ephemer:
Thank you for this change!

PR #281, opened 2025-01-15
Support JS function calls with up to 7 arguments in Embedded Swift

### LINKED ISSUE #279: Unable to call JS functions with more than 2 arguments in Embedded Swift
When building for Embedded Swift, `JSValue` and `JSObject` don’t support JS function calls with more than 2 arguments. I noticed this when trying to call the [`fillRect()` method](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillRect) on `CanvasRenderingContext2D`. Example: 

```swift
import JavaScriptKit

let document = JSObject.global.document
let canvas = document.getElementById("canvas")
var ctx: JSValue = canvas.getContext("2d")
// 🔴 Error (in Embedded Swift only): cannot call value of non-function type 'JSValue'
_ = ctx.fillRect(20, 20, 200, 100)
```

The following works as a workaround but is ugly:

```swift
_ = ctx.object!.fillRect.function!(this: ctx.object!, arguments: [20, 20, 200, 100])
```

The reason that it doesn't work is:

1. Embedded Swift doesn't support existentials, so the general function invocation mechanism of the form `subscript(dynamicMember name: String) -> ((ConvertibleToJSValue...) -> JSValue)` is not available.

2. Both `JSValue` and `JSObject` only have concrete overloads for up to 2 function call arguments:

    ```swift
    #if hasFeature(Embedded)
    public extension JSValue {
        @_disfavoredOverload
        subscript(dynamicMember name: String) -> (() -> JSValue)
    
        @_disfavoredOverload
        subscript<A0: ConvertibleToJSValue>(dynamicMember name: String) -> ((A0) -> JSValue)
    
        @_disfavoredOverload
        subscript<A0: ConvertibleToJSValue, A1: ConvertibleToJSValue>(dynamicMember name: String) -> ((A0, A1) -> JSValue)
    }
    #endif
    ```

---

Question: Would you be interested in a PR that adds more of these concrete overloads for Embedded Swift? Or do you want to wait for variadic generics support in Embedded Swift to solve this problem "the right way"?

If it's the former, I'm willing to send the PR. In this case, up to how many arguments should I go? Is 4 enough? Or should we go up to, say, 6?

ole:
This fixes #279.

Notes:

- The choice up to how many aguments to support before we can replace everything with a single varadic generics funtion is arbitrary. I chose to go up to 7 arguments because `JSFunction.new` already had some overloads with up to 7 arguments, so this seemed the most consistent.

- I used a different code style for the new overloads where each generic parameter is put on its own line. I'm aware that this doesn't follow the project's code style, but I found the alternative (the full function signature on a single line) pretty much unreadable. Please let me know if you want me to change this.

- I didn't add any tests for the new functionality because the existing test suite isn't built/executed in Embedded Swift mode anyway, so the new code wouldn't even be built. I did a quick test locally to ensure the overloads work. I hope this is OK, but please let me know if you want more. If so, we have to integrate an Embedded Swift build into the testing process.

### DIFF
--- a/Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
+++ b/Sources/JavaScriptKit/FundamentalObjects/JSFunction.swift
@@ -164,7 +164,14 @@ public class JSFunction: JSObject {
 }
 
 #if hasFeature(Embedded)
-// NOTE: once embedded supports variadic generics, we can remove these overloads
+// Overloads of `callAsFunction(ConvertibleToJSValue...) -> JSValue`
+// for 0 through 7 arguments for Embedded Swift.
+//
+// These are required because the `ConvertibleToJSValue...` version is not
+// available in Embedded Swift due to lack of support for existentials.
+//
+// Once Embedded Swift supports parameter packs/variadic generics, we can
+// replace all variants with a single method each that takes a generic pack.
 public extension JSFunction {
 
     @discardableResult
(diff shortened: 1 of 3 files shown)


kateinoigakukun (approved):
Seems good to me

PR #306, opened 2025-03-18
Make `JSObject.id` getter inlinable

kateinoigakukun:
This change makes `JSObject.id` getter inlinable. Those who use the API to avoid dynamic member lookup overhead can now benefit from the inlinable getter.


fjtrujy (approved):
👏

PR #342, opened 2025-04-12
Remove UMD build of JS runtime library

kateinoigakukun:
We always use ESM, so we don't need to generate UMD runtime.js anymore.


kateinoigakukun:
Hmm, we still use it for benchmarks...

PR #418, opened 2025-08-21
BridgeJS: Enum export support for case / raw / empty enums

krodak:
## Introduction
This PR adds basic enum export support to the BridgeJS plugin.

### Supported Enum Types
- Case Enums
- Raw Value Enums
- Namespace Enums - empty enums used for organizing types into hierarchical namespaces

### Enum Output Styles
New `@JS(enumStyle: .tsEnum)` parameter allows choosing between:
- **`.const` (default)**: Generates const objects with union types for maximum compatibility
- **`.tsEnum`**: Generates native TypeScript enum declarations (available for case enums and raw value enums with String or numeric raw types only)

### Case Enum with Both Styles
```swift
@JS enum Direction {
    case north, south, east, west
}

@JS(enumStyle: .tsEnum) enum TSDirection {
    case north, south, east, west
}
```

Generated TypeScript:
```typescript
// Const object style (default)
const Direction: {
    readonly North: 0;
    readonly South: 1;
    readonly East: 2;
    readonly West: 3;
};
type Direction = typeof Direction[keyof typeof Direction];

// If @JS(enumStyle: .tsEnum) would be defined
enum TSDirection {
    North = 0,
    South = 1,
    East = 2,
    West = 3,
}
```

### Raw Value Enum
```swift
@JS enum Theme: String {
    case light = "light"
    case dark = "dark"
    case auto = "auto"
}
```

```typescript
// Const object style (default)
const Theme: {
    readonly Light: "light";
    readonly Dark: "dark";
    readonly Auto: "auto";
};
type Theme = typeof Theme[keyof typeof Theme];

// If @JS(enumStyle: .tsEnum) would be defined
enum Theme {
    Light = "light",
    Dark = "dark",
    Auto = "auto",
}
```
### Namespace Enum
```swift
@JS enum Networking {
    @JS enum API {
        @JS enum Method {
            case get
            case post
            case put
            case delete
        }
        // Invalid to declare @JS(namespace) here as it would be conflicting with nesting
        @JS class HTTPServer {
            @JS init() {}
            @JS func call(_ method: Method)
        }
    }
}
// Top level enums can use namespace
@JS(namespace: "Networking.APIV2")
enum Internal {
    @JS enum SupportedMethod {
        case get
        case post
    }
    
    @JS class TestServer {
        @JS init() {}
        @JS func call(_ method: SupportedMethod)
    }
}
```

```typescript
namespace Networking {
        namespace API {
            class HTTPServer {
                constructor();
                call(method: Networking.API.Method): void;
            }
            const Method: {
                readonly Get: 0;
                readonly Post: 1;
                readonly Put: 2;
                readonly Delete: 3;
            };
            type Method = typeof Method[keyof typeof Method];
        }
        namespace APIV2 {
            namespace Internal {
                class TestServer {
                    constructor();
                    call(method: Internal.SupportedMethod): void;
                }
                const SupportedMethod: {
                    readonly Get: 0;
                    readonly Post: 1;
                };
                type SupportedMethod = typeof SupportedMethod[keyof typeof SupportedMethod];
            }
        }
    }
```

## What's Not Supported
- Associated value enums (will be tackled as next step)
- Computed properties or methods within enums

## Testing
- `EnumCase.swift` - Case enum tests with both output styles
- `EnumRawType.swift` - Raw value enum tests for all supported types
- `EnumNamespace.swift` - Namespace enums with mixing same namespace as enum and `@JS(namespace)`

## Docs
Docs updated with example for each enum type, new macro parameter and basic relevant usage guide.


Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/EnumCase.Export.d.ts
```diff
@@ -0,0 +1,44 @@
+// NOTICE: This is auto-generated code by BridgeJS from JavaScriptKit,
+// DO NOT EDIT.
+//
+// To update this file, just rebuild your project or run
+// `swift package bridge-js`.
+
+export const Direction: {
```
kateinoigakukun, line 7: I think it makes sense to declare constant enum type definitions at the top-level scope instead of exporting them from `Exports` since they don't require access to wasm instance.

I think we have a few things to fix here:
1. Generate `export const` definition in the corresponding `.js` file as well as TS and bundlers expect it.
2. Remove the enum definition from exports of `createInstantiator` result in `.js` side

kateinoigakukun (commented):
Overall approach looks good to me!

Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/EnumCase.Export.d.ts
krodak: Thanks for feedback, this definitely makes more sense, addressed this issue and added relevant runtime tests for all types of added enums, will remember about those for future functionalities 🙇🏼‍♂️

PR #432, opened 2025-08-26
added `import _Concurrency` in JSPromise

### LINKED ISSUE #431: Embedded SDK built: `Cannot find 'Task' in scope`
the recent async additions in JSPromise causes the embedded build to break (with 6.2 swift.org wasm sdk)

An explicit `import _Concurrency` would fix the build.
However, I am not 100% sure if
a) it is intentional that _Concurrency must be imported, and
b) if it is good to throw a Task in there without any gates or similar (eg: without the JS executor installed)

```
/home/runner/work/elementary-dom/elementary-dom/Examples/Swiftle/.build/checkouts/JavaScriptKit/Sources/JavaScriptKit/BasicObjects/JSPromise.swift:95:13: error: cannot find 'Task' in scope
 93 |             }
 94 |             let context = Context(resolver: resolver, body: body)
 95 |             Task {
    |             `- error: cannot find 'Task' in scope
 96 |                 do throws(JSException) {
 97 |                     let result = try await context.body()
```

sliemeobn:
should fix #431 

ideally this should be caught in CI

### DIFF
--- a/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
+++ b/Sources/JavaScriptKit/BasicObjects/JSPromise.swift
@@ -1,3 +1,7 @@
+#if hasFeature(Embedded) && os(WASI)
+import _Concurrency
+#endif
+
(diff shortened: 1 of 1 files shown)


kateinoigakukun (commented):
Thanks!

PR #433, opened 2025-08-26
BridgeJS: Fix nested type resolution for MemberTypeSyntax (e.g., Networking.API.Method)

krodak:
## Issue 
Currently for functions that would be using nested types like:

```swift
@JS func echoNetworkingAPIMethod(_ method: Networking.API.Method) -> Networking.API.Method { … }
```

We'd get following error:
```swift
error: Unsupported type: Networking.API.Method
Hint: Only primitive types and types defined in the same module are allowed
```

Issue is that currently `lookupType(for` only handles `IdentifierTypeSyntax`, hence `MemberTypeSyntax` like `Networking.API.Method` would not be resolved to a valid declaration.
 
## Fix
Centralized type resolution in `TypeDeclResolver` with a new `resolve(_: TypeSyntax)` API and used it in `ExportSwift` to map nested enums/classes/actors to `BridgeType`. This should help with nested types resolution, while exported ABI remains unchanged for already-supported types.
New tests were added to `prelude.mjs` to confirm the fix.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -689,47 +689,50 @@ public class ExportSwift {
         if let primitive = BridgeType(swiftType: type.trimmedDescription) {
             return primitive
         }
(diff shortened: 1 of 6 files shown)


kateinoigakukun (approved):
Thanks!

PR #441, opened 2025-09-01
BridgeJS: Migrate to IntrinsicJSFragment for enums

krodak:
## Introduction
Moving most of the code related to generate helper functions for custom lower / raise of associated values in enums to dedicated `IntrinsicJSFragment` and moving to `CodeFragmentPrinter`  instead of "manual" code generation management.
Similar update done for `simple`, `raw` type enums and namespaces.

## Notes
I've also started migrating some code from `BridgeJSLink` to use fragment printer, but I'll post another PR for clarity, as it's not related to enums and I'd rather keep PRs smaller when possible (which is harder to do when adding new functionality with tests)


kateinoigakukun (approved):
Thanks!

PR #442, opened 2025-09-02
[NFC] BridgeJS: Migrate to code fragment printer

krodak:
## Introduction
Migrating from "manual" code generation with the use of `.indent(count`, to using `CodeFragmentPrinter` within `BridgeJSLnk`.

## Notes
- split content of `link()` method as it was becoming a bit large and having lots of moving part as a result of adding new conditional sections for enums, etc
- separated TS and JS code generation within `link()` and encapsulated into individual functions for clarity
- as a side effect removed some empty lines from generated files which were mostly there due to conditional content that was inserted into """ and resulted in empty line if not present
- abstracted vars from `link()` into `LinkData` for clarity
- added some comments for JS / TS code generation to make it a bit easier to identify given segments of JS / TS generated files
- cleaned up unused old version of `renderGlobalNamespace`

Initially was biased to not split `let outputJs` as large """ seemed convenient, but eventually decided on refactoring this as well, as with more functionality added, it would become even more bloated 🙏🏻

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
@@ -65,10 +65,7 @@ struct BridgeJSLink {
         }
         """
 
-    let sharedEnumHelpersJs = """
-        """
-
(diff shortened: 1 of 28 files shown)


kateinoigakukun (approved):
Thanks!

PR #444, opened 2025-09-10
BridgeJS: Swift -> JS Optionals support

krodak:
## Introduction
This PR adds support for Swift Optionals when exporting Swift funcionality to JS / TS using BridgeJS plugin. 

## Overview
- Swift Optionals are represented by `T | null` union, as `null` explicitly represents intentional absent of value, which aligns with Swift's nil.
- for each parameter additional "isSome" flag is used to represent optionality. 
- new shared `extension Optionals` were added to `BridgeJSIntrinsics` to keep generated Swift glue code sparse and reuse existing patterns. This required new protocols for case and associated enum value types, which 
- for support of return values, new bjs module methods were added for JS lowering `_swift_js_return_optional_<type>`
- support for different Optional syntax was added (including typealiasing), so all of the below should work identically:

```swift
@JS func test(name: String?) -> String?
@JS func test(name: Optional<String>) -> Optional<String>
@JS func test(name: Swift.Optional<String>) -> Swift.Optional<String>
@JS func test(name: Swift.Optional<  String >) -> Swift.Optional< String. >
typealias OptionalName = String?
@JS func test(name: OptionalName) -> OptionalName
```

## Examples
```swift
@JS class UserService {
    var userId: Int? 
    func findUser(id: Int) -> User?
    func process(data: String?) -> String
}
```

Generated TypeScript:
```typescript
export interface UserService extends SwiftHeapObject {
    userId: number | null;
    findUser(id: number | null): User | null;
    process(data: string | null): string;
}
```

## Testing
Added tests for different scenarios for all supported optional data types, including parameters, properties and return value usage.

## Documentation
Extended current documentation with new `Exporting-Swift-Optional.md`


kateinoigakukun (approved):
Great job 👏

PR #448, opened 2025-09-19
BridgeJS: Adding support for static / class properties & functions

krodak:
## Introduction
This PR adds support for Swift static / class functions and properties when exporting Swift funcionality to JS / TS using BridgeJS plugin.

## Overview
Both static / class functions are translated to `static` in JS.
Everything beside willSet / didSet is supported similarly to instance properties.
Static functions in namespaced enums are translated to `globalThis` namespaced functions.
Examples can be found in added documentation. 

## Additional changes
- Abstracted ABI naming
- Simplified and tidied up some parts of namespacing
- Fixed issue with nested namespaces - in case of nested namespaces, some elements like functions and properties would be gathered only from most-nested enum

## Resolves
- #435 
- #438 

## Testing
Added tests for different scenarios for all supported optional data types, including parameters, properties and return value usage.

## Documentation
Extended current documentation with new
- [Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Exporting-Swift/Exporting-Swift-Static-Functionss.md](https://github.com/swiftwasm/JavaScriptKit/pull/448/files#diff-378f7dd6b00e68c28e6a1dc70260f09a3f80c626787adb96968a356324b01542)
- [Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Exporting-Swift/Exporting-Swift-Static-Properties.md](https://github.com/swiftwasm/JavaScriptKit/pull/448/files#diff-beb3bdd7dfbf20cef2221113c9b13fce3254ed5f1e25aa0143b49f3fe3acb0aa)


Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/StaticFunctions.Export.js
```diff
@@ -0,0 +1,292 @@
...
+            if (typeof globalThis.Utils === 'undefined') {
+                globalThis.Utils = {};
+            }
+            if (typeof globalThis.Utils.String === 'undefined') {
+                globalThis.Utils.String = {};
+            }
+            Calculator.square = function(value) {
+                const ret = instance.exports.bjs_Calculator_static_square(value);
+                return ret;
+            };
```
kateinoigakukun, line 268: I think BridgeJS should support multiple instantiations as long as the global namespace feature is not used in principle, so I'd like to keep any APIs accessing an instance in the `Exports` set returned by an instantiation.

So I expect the generated TS interface would have something like below (we need a better naming convention for the constant value type and the type holding static methods though 😅):

```ts
export const Calculator: {
    readonly Scientific: 0;
    readonly Basic: 1;
};
export type CalculatorConst = typeof Calculator[keyof typeof Calculator];

export type Exports = {
    Calculator: {
        square(value: number): number;
    }
}
```

kateinoigakukun (commented):
Overall, looks good to me, modulo the enum static properties and methods APIs. Thanks!

Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/StaticFunctions.Export.js
krodak: Sounds reasonable 👌 I’ll update PR after next week though, so hang in there 🙏

What about ‚namespace enums’, are you fine with globalThis namespaces properties and functions as implemented now?

kateinoigakukun: Thanks!

For namespace enums, I have been thinking more about that feature, but we might need to split the "namespacing" feature and the "exposing to global" feature for better consistency.

```swift
@JS enum Utils {
    @JS enum String {
        @JS static func uppercase(_ text: String) -> String {
            return text.uppercased()
        }
    }
}
@JS(namespace: "Networking.APIV2")
enum Internal {
    @JS enum SupportedMethod {
        case get
        case post
    }
}
```

I expect the above code will generate the following interface by default:

```ts
export type Exports = {
    Utils: {
        String: {
            uppercase(text: string): string;
        }
    },
    Networking: {
        APIV2 {
            Internal: {
                SupportedMethod: {
                    readonly Get: 0;
                    readonly Post: 1;
                },
            }
        }
    }
}
```

Then if we set `{ "exposeToGlobal": true }` in `bridge-js.config.json`, then it will add:

```ts
declare global {
    namespace Utils {
        namespace String {
            uppercase(text: string): string;
        }
    }
    namespace APIV2 {
        namespace Internal {
            const SupportedMethod: {
                readonly Get: 0;
                readonly Post: 1;
            };
            type SupportedMethod = typeof SupportedMethod[keyof typeof SupportedMethod];
        }
    }
}
```

But I think that change would be out of scope in this PR, so I'm fine with keeping it as is for now.

PR #453, opened 2025-10-02
[BridgeJS] Support default values

krodak:
## Introduction

This PR adds support for default parameter values when exporting Swift functions and constructors to JavaScript/TypeScript using the BridgeJS plugin.

Resolves: https://github.com/swiftwasm/JavaScriptKit/issues/452

## Overview
* Default parameters are represented as optional parameters (`param?: type`) in TypeScript, using `undefined` to trigger default value application, while not conflict with support for optionals
* Generated JavaScript uses native default parameter syntax
* JSDoc comments automatically document default values in TypeScript definitions
* Supported defaults: literals, `nil`, enum cases, and object initialization (e.g. `MyClass("arg", 42)`)
* There are some additional checks for unsupported expression types used as default values, so those should result in clear error (checks for method calls, closures, unsupported types, etc)

## Examples

```swift
@JS public func greet(name: String = "World", enthusiastic: Bool = false) -> String {
    return enthusiastic ? "Hello, \(name)!" : "Hello, \(name)"
}

@JS class Config {
    @JS var name: String
    @JS var timeout: Int
    
    @JS init(name: String = "default", timeout: Int = 30) {
        self.name = name
        self.timeout = timeout
    }
}
```

Generated JavaScript:

```javascript
export function greet(name = "World", enthusiastic = false) {
    ...
}

export class Config extends SwiftHeapObject {
    constructor(name = "default", timeout = 30) {
        ...
    }
}
```

Generated TypeScript:

```typescript
export type Exports = {
    /**
     * @param name - Optional parameter (default: "World")
     * @param enthusiastic - Optional parameter (default: false)
     */
    greet(name?: string, enthusiastic?: boolean): string;
}

export interface Config extends SwiftHeapObject {
    /**
     * @param name - Optional parameter (default: "default")
     * @param timeout - Optional parameter (default: 30)
     */
    new(name?: string, timeout?: number): Config;
}
```

## Testing
Added tests for different scenarios covering all supported default value types, including parameters, constructors, negative numbers, optionals, enums, and object initialization.

## Documentation
Added documentation in `Exporting-Swift-Default-Parameters.md`.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -243,48 +243,18 @@ public class ExportSwift {
                 return .null
             }
 
-            if let stringLiteral = expr.as(StringLiteralExprSyntax.self),
-                let segment = stringLiteral.segments.first?.as(StringSegmentSyntax.self),
-                type.matches(against: .string)
-            {
-                return .string(segment.content.text)
-            }
-
-            if let boolLiteral = expr.as(BooleanLiteralExprSyntax.self),
-                type.matches(against: .bool)
-            {
-                return .bool(boolLiteral.literal.text == "true")
-            }
-
-            if let intLiteral = expr.as(IntegerLiteralExprSyntax.self),
-                let intValue = Int(intLiteral.literal.text),
-                type.matches(against: .int)
-            {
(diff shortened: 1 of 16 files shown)


kateinoigakukun (approved):
Thank you for the detailed explanation! If you have a lot of existing code with default parameter values, I think it makes sense to merge this so as not to block the adoption.

PR #455, opened 2025-10-17
BridgeJS: Fix support for negative enum values for raw value types

krodak:
## Description
Fixing issue where using negative value in enum of numeric raw type would result in value not being process and fallback being used


kateinoigakukun (commented):
Would you mind updating runtime test as well?

PR #457, opened 2025-10-18
Add `--default-platform` option to `package js`

t089:
Currently `swift package js` only generates code for the "browser" platform.

This PR adds a `--platform node|browser [default: browser]` argument that lets you control this behaviour. In case of `node` it will generate a `init` function that calls `defaultNodeSetup` instead of `defaultBrowserSetup`.


Plugins/PackageToJS/Sources/PackageToJS.swift
```diff
@@ -8,6 +8,8 @@ struct PackageToJS {
         var configuration: String?
         /// Name of the package (default: lowercased Package.swift name)
         var packageName: String?
+        /// Target platform for the generated JavaScript (default: browser)
+        var platform: String?
```
MaxDesiatov: Could this become an `enum Platform: CaseIterable`, so then you would rely on `.allCases` to dynamically compute help string for all available platforms in the help output you've added below for this new option?

kateinoigakukun (commented):
Thank you for putting your effort here! Overall it makes sense to me!

Let me leave a few feedbacks before mering:
- I'd like to keep the generated JS package to be runtime independent as much as possible, so it might be better to name the option as `--default-platform`, indicating that only the default entrypoint `index.js` is platform dependent, and rest of the modules are still platform independent. The clarification would be partuculary helpful for library author publishing a platform-agnostic JS package containing the JSKit genrated code.
- I'm not a big fan of adding `@types/node` dependency here because we use only small surface of Node.js API, and most of the usage is not visible from users (except for `Worker` type). I think we can loosen type checks in the Node.js for the sake of the simplicity.

If that sounds reasonable, I can make changes on the top of your branch and merge this PR :)

PR #459, opened 2025-10-21
PackageToJS: Expose Playwright Launch Options in test harness

kateinoigakukun:
This allows a bit more flexibility to developers to customize test runner configurations


fjtrujy (approved):
Amazing!

Issue #477, opened 2025-12-15
JavaScript runtime portability


sliemeobn:
I am working on a smooth vite-based "hot-reload" + build experience for an ElementaryUI starter template. I have it working well now, but the topic of how to bootstrap the javascript parts is still messy. Currently, the only way to get to the required JavaScriptKit runtime imports reliably is by building through the `js` plugin. I think the `js` command plugin is a great choice for many cases, but it has issues.

For my "web app in the browser" use case:
+ debug rebuild on file-change takes a lot longer (on my machine ~2 s vs ~0.9 s "raw" swift build - most of it probably just SwiftPM waking up)
+ the js plugin simply checks a lot of things that I do not need every time
+ the wasi-shim dependency is in a weird place (cdn is no good for "real" stuff, and depending on a local package that does not exist until after build is also weird)
+ I don't love that I need a matching top-level JavaScriptKit dependency (especially since we are still in 0.x.x versions)

For all these reasons I decided to base my "swift wasm" vite plugin on "raw" SwiftPM build commands. With the SwiftWasm SDKs the build is really nothing I need a plugin for (only the linker bits for `exec-model=reactor` and main export are a bit rough, but very doable).

I currently see two and a half issues:
+ How do I get the javascript runtime reliably?
+ How do I deal with the bridge-js imports reliably?
+ How to best deal with the WASI-shim setup (not a blocker really)

Here is my annotated WIP-version of the `index.ts` that would bootstrap an ElementaryUI app:
```ts
// index.ts
import { WASI, OpenFile, File, ConsoleStdout } from "@bjorn3/browser_wasi_shim";

// PROBLEM: typing bug makes this non-compatible with esbuild/rolldown
// - "export { SwiftRuntimeThreadChannel };",
// + "export type { SwiftRuntimeThreadChannel };"
import { SwiftRuntime } from "./.build/checkouts/JavaScriptKit/Runtime/src/index.ts";

// PROBLEM: these bjs imports live nowhere, only in the js plugin templates 
import { bjs } from "./javascriptkit.ts";

import appInit from "./Package.swift"; // this is the cool part that the vite plugin does

// JavaScriptKit Swift runtime
const swiftRuntime = new SwiftRuntime();

// HALF-PROLEM: do we ever need to customize this, or should we just vendor a pre-configured version of this?
// Basic WASI setup
const wasi = new WASI(
  [],
  [],
  [
    new OpenFile(new File([])),
[...]

PR #478, opened 2025-12-15
Runtime portability improvements

sliemeobn:
as discussed in #477

This PR
+ makes sure the Runtime remains compatible with "isolatedModules" (ie: esbuild, rollup, ...)
+ move the JSBridge stubs out of the plugin template for better reusability

I think these changes are good even when the whole "how to best bundle/deploy/build/package" is not fully clear yet.

### DIFF
--- a/Runtime/tsconfig.json
+++ b/Runtime/tsconfig.json
@@ -8,8 +8,7 @@
         "strict": true,
         "target": "es2017",
         "lib": ["es2020", "DOM", "ESNext.WeakRef"],
-        "skipLibCheck": true,
-        "isolatedModules": true
+        "skipLibCheck": true
     },
     "include": ["src/**/*"],
     "exclude": ["node_modules"]


sliemeobn:
unfortunately this tsconfig setting `"isolatedModules": true` messes up the codegen for `export const enum Kind`
I reverted it for now, but maybe we should add CI checks for the Runtime "buildability" somehow?

kateinoigakukun:
I still do not want to expose the Runtime code as an API because maintaining a stable interface at that layer requires much effort.

kateinoigakukun:
Thanks for summarizing it :pray: I support the first two points, but still not very sure about the third one. I don't think the third one is independent from https://github.com/swiftwasm/JavaScriptKit/issues/477's direction, so I'm a bit confused.

At least, exposing every bjs intrinsic import function names as type declaration increases the number of files we need to edit to add intrinsic, and also I personally don't want to consider the import function list being a public API.

How about exposing `createInstantiator` instead? It allows us to keep the BridgeJS stubs as implementation details while giving you the flexibility to satisfy imports with custom bootstrap code outside of PackageJS.

sliemeobn:
> How about exposing createInstantiator instead? It allows us to keep the BridgeJS stubs as implementation details while giving you the flexibility to satisfy imports with custom bootstrap code outside of PackageJS.

I like it, this might be the best choice. I'll update the PR, but I first want to find the time to wrap my head around the API shape, and how that might all play out with composing multiple libraries and multiple JS imports/snippets.

PR #481, opened 2025-12-23
BridgeJS: Struct and struct vs classes benchmarks

krodak:
## Overview

This PR adds performance benchmarks for Swift structs, similar to the existing enum benchmarks. 
Also added some class equivalents for direct struct vs class comparison.

## Benchmark Coverage

  **Structs:**
  - `SimpleStruct` - 5 fields (String, Int, Bool, Float, Double)
  - `Address` - 3 fields (strings + int)
  - `Person` - nested struct with optional
  - `ComplexStruct` - 6 fields

  **Classes (for comparison with structs):**
  - `SimpleClass` - equivalent to SimpleStruct
  - `AddressClass` - equivalent to Address

 ## Results

 ### Struct Benchmarks

`node run.js --adaptive --filter=StructRoundtrip`

CleanShot 2025-12-23 at 19 06 44@2x

  ### Struct vs Class Comparison

 `node run.js --adaptive '--filter=/Simple|Address/'`

CleanShot 2025-12-23 at 19 04 57@2x

### DIFF
--- a/Benchmarks/Sources/Benchmarks.swift
+++ b/Benchmarks/Sources/Benchmarks.swift
@@ -111,6 +111,136 @@ enum ComplexResult {
     }
 }
 
+// MARK: - Struct Performance Tests
+
+@JS struct SimpleStruct {
+    var name: String
+    var count: Int
+    var flag: Bool
+    var rate: Float
+    var precise: Double
+}
+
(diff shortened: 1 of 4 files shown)


kateinoigakukun (approved):
Thanks!

PR #482, opened 2025-12-24
[NFC] BridgeJS: Codegen code organisation and minor cleanups

krodak:
## Overview 

This PR extracts some codegen code from ExportSwift, as it became a bit large due to new types introduction. Some additional minor cleanup introduced along the way.

## Technical changes

- extracted specialized code generators into `ExportSwiftCodegens.swift` with updated and a bit more consistent naming
- added protocols codegen
- unified thunk body rendering (`renderFunctionBody(into:returnExpr:)` helper)
- shared property rendering for instance and static ones
- removed some unused parameters
 
## Testing 
I've fixed testing regression I've introduced [here](https://github.com/swiftwasm/JavaScriptKit/pull/480/changes#diff-fd034a6b3245c6475995a7497846e184065bf873d332a0bfbe4f2ad476356e74L1376) when I was trying to run some tests on macOS with Swift 6.2.1 and disabled async tests, sorry for that 🙇🏻  🙏🏻


kateinoigakukun (approved):
Thanks!

PR #503, opened 2026-01-21
chore: Minor updates to BridgeJS documentation

krodak:
## Overview

Some more updates to align existing BridgeJS documentation.
- aligned styling, structure for Exporting to Swift articles, to make it more consistent for the reader. Included some technical explanation in dedicated "How it works" section, so it's a bit more obvious if we copy, use references, etc
- cleaned up a bit technical bridgeJS readme, let me know if that's fine. I remember this being my introduction point for the project, so I thought it could use some attention, might be still useful for potential contributors or folks working on other related projects 🙏🏻


Plugins/BridgeJS/README.md
```diff
@@ -59,27 +60,53 @@ graph LR
...
+| `UInt16`      | `number`        | `i32`          |
+| `Int32`       | `number`        | `i32`          |
+| `UInt32`      | `number`        | `i32`          |
+| `Int64`       | `bigint`        | `i64`          |
+| `UInt64`      | `bigint`        | `i64`          |
+| `Float`       | `number`        | `f32`          |
+| `Double`      | `number`        | `f64`          |
+| `Bool`        | `boolean`       | `i32`          |
+| `Void`        | `void`          | -              |
+| `String`      | `string`        | `i32` (pointer)|
```
kateinoigakukun:
```suggestion
```

Let's not include String here as its convention is not very "primitive"-ish 😅

kateinoigakukun (approved):
Thanks!

PR #511, opened 2026-01-22
Test building `BridgeJSTool` on macOS on CI

MaxDesiatov:
This exposes build errors on macOS that aren't reproducible on Linux.

Lowered SwiftSyntax dependency range to 601 due to following error with 602:

```
ImportTS.swift:736:16: error: no exact matches in call to initializer 
734 |     /// Builds the standard @_expose and @_cdecl attributes for WebAssembly-exposed functions
735 |     static func buildExposeAttributes(abiName: String) -> AttributeListSyntax {
736 |         return AttributeListSyntax {
    |                `- error: no exact matches in call to initializer 
737 |             AttributeSyntax(
738 |                 attributeName: IdentifierTypeSyntax(name: .identifier("_expose")),
```

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -37,7 +37,7 @@ let package = Package(
         .plugin(name: "BridgeJSCommandPlugin", targets: ["BridgeJSCommandPlugin"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/swiftlang/swift-syntax", "600.0.0"..<"603.0.0")
+        .package(url: "https://github.com/swiftlang/swift-syntax", "600.0.0"..<"601.0.0")
     ],
     targets: [
         .target(


MaxDesiatov:
@kateinoigakukun @krodak this looks like something that should be exposed on Linux as well?

```
ImportTS.swift:736:16: error: no exact matches in call to initializer 
734 |     /// Builds the standard @_expose and @_cdecl attributes for WebAssembly-exposed functions
735 |     static func buildExposeAttributes(abiName: String) -> AttributeListSyntax {
736 |         return AttributeListSyntax {
    |                `- error: no exact matches in call to initializer 
737 |             AttributeSyntax(
738 |                 attributeName: IdentifierTypeSyntax(name: .identifier("_expose")),
```

PR #560, opened 2026-02-03
BridgeJS: Remove `@JS protocol` in favor of `@JSClass struct`

kateinoigakukun:
## Motivation

The main reason we introduced `@JS protocol` was just because we didn't have a Swift-level notation to mix import and export decls but we today have `@JSClass struct`, so we no longer need it.

## Summary
- Drop protocol collection and wrapper generation paths; steer diagnostics to `@JSClass` structs instead.
- Replace protocol fixtures/docs/runtime coverage with JSClass-based equivalents and regenerate outputs.
- Update runtime harness to exercise the JSClass interface path previously covered by protocols.

### DIFF
--- a/Tests/BridgeJSRuntimeTests/ExportAPITests.swift
+++ b/Tests/BridgeJSRuntimeTests/ExportAPITests.swift
@@ -954,6 +954,86 @@ enum APIOptionalResult {
     }
 }
 
+// MARK: - JSClass Interface Tests
+
+@JSClass struct DataProcessor {
+    @JSGetter var count: Int
+    @JSGetter var name: String
+    @JSGetter var tag: String
+
+    @JSFunction func increment(by amount: Int) throws(JSException)
+    @JSFunction func getValue() throws(JSException) -> Int
+    @JSFunction func setLabel(_ labelPrefix: String, _ labelSuffix: String) throws(JSException)
+    @JSFunction func getLabel() throws(JSException) -> String
+    @JSFunction func isEven() throws(JSException) -> Bool
+    @JSFunction func processNote(_ note: String) throws(JSException) -> String
+    @JSFunction func createNote(_ prefix: String) throws(JSException) -> String
+    @JSFunction func handleMessage(_ message: String) throws(JSException)
+    @JSFunction func getMessage() throws(JSException) -> String
+
+    @JSFunction func setTag(_ tag: String) throws(JSException)
+}
+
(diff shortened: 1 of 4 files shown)


krodak:
Thanks for the cleanup work here! Before this lands, I wanted to discuss some use cases we have in our codebase that rely on `@JS protocol` for **export** scenarios. I might be missing something about how `@JSClass struct` covers these, so please let me know if there's an alternative approach.

### Our example use case: common interface for multiple exported classes

We have 50+ wrapper classes that all need to be usable with a graph API:

```swift
@JS protocol QuantumNodeProvider {
    var quantumObjectBridged: QuantumObjectBridged { get }
}

@JS class AdjacencyBridged: QuantumNodeProvider { ... }
@JS class SurfaceBridged: QuantumNodeProvider { ... }
@JS class BuildingBridged: QuantumNodeProvider { ... }
// ... 50+ more

// Function accepts ANY wrapper
@JS static func createGraph(
    root: QuantumNodeProvider,
    nodes: [QuantumNodeProvider]?
) -> RootedQuantumGraphV2
```

On the TypeScript side, this generates an interface that any wrapper satisfies:

```typescript
const adj = new AdjacencyBridged(...);
const surf = new SurfaceBridged(...);
const graph = RootedQuantumGraphV2.create(adj, [surf]); // Mixed types work
```

## Questions

1. **How would we express "accepts any of these 50+ classes" without `@JS protocol`?** With `@JSClass struct`, I don't see how to define a shared interface on the export side.

2. **Is there a pattern for protocol-typed return values?** E.g., `@JS func createNode() -> QuantumNodeProvider` where Swift decides which concrete type to return.

3. **Delegate/callback pattern** - we also have cases where JS implements an interface and passes it to Swift. Is there an alternative path for that?

We're in the middle of migrating from manual `JSClosure` bindings to `BridgeJS`. Our existing architecture heavily uses protocols for polymorphism .

For context, we added `@JS protocol` support back in October (#456) specifically to enable this migration path. Our plan was to use protocol conformance to gradually port our 50+ model wrappers while keeping them interoperable through shared interfaces like `QuantumNodeProvider`.

We could theoretically refactor to avoid protocols entirely, but that would mean changing direction on the migration approach we've been building towards. If there's a way to keep protocol support (or an equivalent mechanism for the export side), that would help us a lot. Happy to discuss alternatives if there are concerns with the current implementation.

PR #563, opened 2026-02-03
BridgeJS: Enforce throws(JSException) on @JS protocol methods

krodak:
## Overview
- Require `throws(JSException)` on `@JS protocol` methods, matching the import side (`@JSFunction`) behavior
- Fix generated protocol wrapper code to propagate throws effect
- Updated tests and documentation

### Test updates

- `Protocol.swift` test input: Added `throws(JSException)` to all protocol methods
- `ExportAPITests.swift`: Updated `DataProcessor` protocol methods with throws
- Regenerated snapshots

### Documentation

- Updated `Exporting-Swift-Protocols.md` with throws requirement and examples

## Motivation
This partially addresses [the inconsistency](https://github.com/swiftwasm/JavaScriptKit/pull/560#issuecomment-3840061515) where `@JSFunction` (import side) requires `throws(JSException)` but `@JS protocol` methods (export side) didn't. 

Properties will be covered in separate PR 🙏🏻

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -1346,7 +1346,7 @@ struct ProtocolCodegen {
             let builder = ImportTS.CallJSEmission(
                 moduleName: moduleName,
                 abiName: "_extern_\(method.name)",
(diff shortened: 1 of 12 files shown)


kateinoigakukun (approved):
Thanks!

PR #576, opened 2026-02-04
BridgeJS: Support enum types as associated values in exported enums

krodak:
## Overview

This PR adds support for case enums and raw value enums (including their optional variants) as associated value payloads in exported enums. Previously only primitive types (`String`, `Int`, `Float`, `Double`, `Bool`) were supported.


krodak:
Closing in favor of: https://github.com/swiftwasm/JavaScriptKit/pull/579

Issue #536, opened 2026-01-28
[BridgeJS] Swift closures are leaked when we bring them to JS world


kateinoigakukun:
The following cases will leak closure instances forever

```swift
@JSFunction func takeClosure(_ closure: @escaping () -> Void)
@JS func makeClosure() -> () -> Void
```

Issue #541, opened 2026-01-28
Looking for API ideas: Typed closure handle for BridgeJS


kateinoigakukun:
## Summary

BridgeJS needs a typed closure handle type that preserves object identity across the Swift/JS bridge. We're looking for community input on naming.

## Background

JavaScriptKit already has `JSClosure` for untyped closures:

```swift
let closure = JSClosure { args -> JSValue in
    // ...
}
```

For BridgeJS, we need a **typed** variant with a generic parameter for the function signature. This enables type-safe declarations like:

```swift
@JSFunction func addEventListener(_ type: String, _ listener: (MessageEvent) -> Void)
```

The typed closure handle is essential for APIs like `addEventListener`/`removeEventListener` that require stable object identity, passing the same closure reference to both calls.

## Related issues

- https://github.com/swiftwasm/JavaScriptKit/issues/536
  - We need the explicit handle to allow manually releasing JS-passed closures 
- https://github.com/swiftwasm/JavaScriptKit/issues/540

## Constraints

1. Must coexist with `JSClosure`: we can't simply make `JSClosure<T>` generic (breaking change)
2. Explicit conversion required: Swift doesn't support implicit conversion from closure literals, so users will write the type name frequently

### 1. Nested type: `JSClosure.Typed<T>`

```swift
let handler = JSClosure.Typed<(MessageEvent) -> Void> { event in
    print(event.data)
}
```

- Clear relationship to `JSClosure`
- Discoverable via autocomplete
- Verbose (15 characters)

## Usage Context

```swift
// Declaration (generated by BridgeJS)
@JSInterface class EventTarget {
    func addEventListener(_ type: String, _ listener: ClosureHandle<(Event) -> Void>)
    func removeEventListener(_ type: String, _ listener: ClosureHandle<(Event) -> Void>)
}

// User code
let handler = ClosureHandle<(MessageEvent) -> Void> { event in
    print(event.data)
}
element.addEventListener("message", handler)
// later...
element.removeEventListener("message", handler)
```

(Replace `ClosureHandle` with the chosen name)

Issue #540, opened 2026-01-28
[BridgeJS] Provide a way to take a JSObject handle of Swift closure passed to JS


kateinoigakukun:
Without that, we cannot use `addEventListener`/`removeEventListener` API properly

PR #578, opened 2026-02-04
BridgeJS: Add `JSTypedClosure` API

kateinoigakukun:
BridgeJS: Add `JSTypedClosure` API

The new API allows managing JS closures converted from Swift closures from Swift side. It allows us to get the underlying `JSObject` (https://github.com/swiftwasm/JavaScriptKit/issues/540) and manage its lifetime manually from Swift.

```swift
@JS func makeIntToInt() throws(JSException) -> JSTypedClosure<(Int) -> Int> {
    return JSTypedClosure { x in
        return x + 1
    }
}

@JSFunction func takeIntToInt(_ transform: JSTypedClosure<(Int) -> Int>) throws(JSException)

let closure = JSTypedClosure<(Int) -> Int> { x in
    return x * 2
}
defer { closure.release() }
try takeIntToInt(closure)
```

Likewise to `JSClosure`, API users are responsible for "manually" releasing the closure when it's no longer needed by calling `release()`. After releasing, the closure becomes unusable and calling it will throw a JS exception (note that we will not segfault even if the closure is called after releasing).

```swift
let closure = JSTypedClosure<(Int) -> Int> { x in
    return x * 2
}
closure.release()
try closure(10)  // "JSException: Attempted to call a released JSTypedClosure created at <file>:<line>"
```

As a belt-and-suspenders measure, the underlying JS closure is also registered with a `FinalizationRegistry` to ensure that the Swift closure box is released when the JS closure is garbage collected, in case the API user
forgets to call `release()`. However, relying on this mechanism is **not recommended** because the timing of garbage collection is non-deterministic and it's not guaranteed that it will happen in a timely manner.

----

On the top of the new API, this commit also fixes memory leak issues of
closures exported to JS.

Close https://github.com/swiftwasm/JavaScriptKit/issues/536 https://github.com/swiftwasm/JavaScriptKit/issues/540 https://github.com/swiftwasm/JavaScriptKit/issues/541

----

## Future TODO

- [ ] Ban usage of `JSTypedClosure` in parameters of exported Swift functions and results of imported JS functions.
- [ ] Allow closure signature to have `throws(JSException)`


krodak (approved):
🙌🏻 🙇🏻

PR #588, opened 2026-02-05
[NFC] BridgeJS: Remove explicit stack storage captures from enum/struct helpers

kateinoigakukun:
After https://github.com/swiftwasm/JavaScriptKit/pull/579, they naturally have access to those storages


krodak (approved):
Thank you, forgot to discuss that, as I wasn't sure what's better convention onJS side, whether it's explicit pass or scope capture is preferred 🙏🏻

Issue #512, opened 2026-01-22
BridgeJS hangs then produces errors for DOM types


MaxDesiatov:
Paste following TypeScript code into PlayBridgeJS in `bridge-js.d.ts` tab (or open this link https://swiftwasm.org/JavaScriptKit/PlayBridgeJS/?code=H4sIAAAAAAAAE6tWKi7PTCtRslLS11fwL0jNU0hIKspMSU%2FVzSrWS9ErKU5QKElMUtJRSikpBipKrSjILypRSCvNSy7JzM9TSE8tcU7MK0ss1qiwUiguKcrMS9e0UvAI8fWBCLvmpOam5pVYK9UCANyqcRdpAAAA&enc=gzip-b64)

```typescript
export function getCanvas(x: string): HTMLCanvasElement;
```

this will make the page hang for long time (a minute or two), then producing an error displayed above the text input that's not actionable:

```
[...]
JavaScriptKit/node_modules/typescript/lib/lib.dom.d.ts:8654:4: warning: Unknown non-nominal type: IDBObjectStore | IDBIndex | IDBCursor
JavaScriptKit/node_modules/typescript/lib/lib.dom.d.ts:377:4: warning: Unknown non-nominal type: boolean | MediaTrackConstraints
JavaScriptKit/node_modules/typescript/lib/lib.dom.d.ts:8545:4: warning: Unknown non-nominal type: IDBValidKey | IDBKeyRange
BridgeJSTool/ImportTS.swift:75: Assertion failed: For now, we require a single parameter to be lowered to a single Wasm core type
[5/8] Import TypeScript API
error: Plugin ended with exit code 1

```

kateinoigakukun:
Perf-tune part 1 https://github.com/swiftwasm/JavaScriptKit/pull/577

PR #617, opened 2026-02-11
BridgeJS: Remove BasicFormat usage in ExportSwift

kateinoigakukun:
Mitigate https://github.com/swiftwasm/JavaScriptKit/issues/512

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -91,8 +91,7 @@ public class ExportSwift {
             }
         }
         return withSpan("Format Export Glue") {
-            let format = BasicFormat()
-            return decls.map { $0.formatted(using: format).description }.joined(separator: "\n\n")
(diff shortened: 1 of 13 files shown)


Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
```diff
@@ -1035,22 +1027,23 @@ struct StackCodegen {
...
-            statements.append(stmt)
+            statements.append(stmt.description)
         }
 
         statements.append("}")
-        statements.append("_swift_js_push_i32(Int32(\(raw: accessor).count))")
-        return statements
+        statements.append("_swift_js_push_i32(Int32(\(accessor).count))")
+        let parsed: CodeBlockItemListSyntax = "\(raw: statements.joined(separator: "\n"))"
+        return Array(parsed)
```
MaxDesiatov: Redundant copy here? Could this return `some Sequence<CodeBlockItemSyntax>` instead of you want it to avoid leaking `CodeBlockItemListSyntax` as a return type?

PR #622, opened 2026-02-12
NFC: BridgeJS: Centralize type info on BridgeType, remove dead cleanup infrastructure

krodak:
# NFC: BridgeJS: Centralize type info on BridgeType, remove dead cleanup infrastructure

## Summary

As recently discussed, this PR centralizes scattered type-specific knowledge as computed properties on `BridgeType`. Codegen reads these properties directly instead of re-deriving the same facts in each switch. Codegen still switches on `BridgeType` cases, grouping them when they share logic.

Also removes the cleanup infrastructure that became dead code after #635 and #636.

64 files changed, +3667/-5050, net -1383 lines.

### BridgeType computed properties (BridgeJSSkeleton.swift)

Previously, type-specific knowledge (ABI shape, lowering method, optional convention, etc.) was scattered across switches in ExportSwift, ImportTS, BridgeJSLink, and JSGlueGen. Each file re-derived the same facts independently. This PR gathers that information into computed properties on `BridgeType` itself, so when adding or modifying a type, there's one place to look.

10 focused properties organized by consumer:

- **ABI Shape** - `wasmParams`, `wasmReturnType`, `optionalConvention`
- **ExportSwift** - `accessorTransform`, `lowerMethod`, `usesStackLifting`
- **ImportTS** - `importParams`, `importReturnType`
- **JSGlueGen** - `nilSentinel`, `optionalScalarKind`

Supporting types: `OptionalConvention`, `NilSentinel`, `OptionalScalarKind`, `AccessorTransform`, `LowerMethod`.

`NilSentinel` captures whether a type has a bit pattern that represents nil without an extra `isSome` flag:

- `jsObject`, `swiftProtocol` - sentinel `0` (object IDs start at 2)
- `swiftHeapObject` - null pointer
- `caseEnum`, `associatedValueEnum` - sentinel `-1` (never a valid case index)

### JSGlueGen helpers (JSGlueGen.swift)

Private helpers that switch on `BridgeType` for JS-side coercion: `isSingleParamScalar()`, `stackLowerCoerce()`, `liftCoerce()`, `lowerCoerce()`, `varHint()`. These consolidate the type-to-JS-coercion mapping that was previously spread across the four direction functions.

### Dead cleanup infrastructure removal

After #635 and #636, no type produces cleanup code. Removed all of it:

- `cleanupCode` printer from `IntrinsicJSFragment.PrintCodeContext`
- `tmpStructCleanups` array and `swift_js_struct_cleanup` intrinsic handler
- `_swift_js_struct_cleanup` extern and wrapper from `BridgeJSIntrinsics.swift`
- Struct helper `lower()` no longer returns `{ cleanup }` - just pushes fields
- Enum helper `lower()` returns caseId directly instead of `{ caseId, cleanup }`
- `init(unsafelyCopying:)` no longer calls struct cleanup
- All optional/array/dictionary wrappers simplified

~1680 lines removed across source and snapshots.

### Compositional optional handling

`optionalLowerParameter` and `optionalLiftParameter` compose T's existing fragment inside an `isSome` conditional instead of switching per type. New types added to `lowerParameter` or `liftParameter` get optional handling for free. Same for struct fields via `structFieldLowerFragment`.

### Other generated code improvements

- **JSValue optional lowering** - `__bjs_jsValueLower` no longer called on null values; the `isSome` guard runs before lowering
- **Struct/enum helper factory collapse** - `() => { return () => ({...}); }` simplified to `() => ({...})`; call-site `()()` becomes `()`
- **Typed defaults in optional lowering** - `0.0` for floats, `flag ? 1 : 0` for bools (previously `0` for everything)
- **Dead code removal** - unused functions in ExportSwift, JSGlueGen, BridgeJSSkeleton

### DIFF
--- a/Benchmarks/Sources/Generated/BridgeJS.swift
+++ b/Benchmarks/Sources/Generated/BridgeJS.swift
@@ -132,10 +132,7 @@ extension SimpleStruct: _BridgedSwiftStruct {
     }
 
     init(unsafelyCopying jsObject: JSObject) {
-        let __bjs_cleanupId = _bjs_struct_lower_SimpleStruct(jsObject.bridgeJSLowerParameter())
-        defer {
-            _swift_js_struct_cleanup(__bjs_cleanupId)
-        }
+        _bjs_struct_lower_SimpleStruct(jsObject.bridgeJSLowerParameter())
         self = Self.bridgeJSStackPop()
     }
 
@@ -148,9 +145,9 @@ extension SimpleStruct: _BridgedSwiftStruct {
 
 #if arch(wasm32)
 @_extern(wasm, module: "bjs", name: "swift_js_struct_lower_SimpleStruct")
-fileprivate func _bjs_struct_lower_SimpleStruct(_ objectId: Int32) -> Int32
+fileprivate func _bjs_struct_lower_SimpleStruct(_ objectId: Int32) -> Void
 #else
-fileprivate func _bjs_struct_lower_SimpleStruct(_ objectId: Int32) -> Int32 {
+fileprivate func _bjs_struct_lower_SimpleStruct(_ objectId: Int32) -> Void {
     fatalError("Only available on WebAssembly")
 }
 #endif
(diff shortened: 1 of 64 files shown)


krodak:
@kateinoigakukun thanks for looking at this, I'll think on your feedback and try to progress in this direction when I can; in parallel I'll look for more simplifications around intrinsics like we are both currently doing 👌🏻

PR #631, opened 2026-02-13
Prevent cross-module inlining of remaining @_extern(wasm) functions

fjtrujy:
## Summary

Extends the fix from #628 (`b3ddd88f`) to all remaining public `@_extern(wasm)` BridgeJS intrinsics that were missing the `@inline(never)` wrapper pattern:

- `_swift_js_throw`
- `_swift_js_init_memory`
- `_swift_js_push_string`
- `_swift_js_push_i32`
- `_swift_js_pop_i32`
- `_swift_js_struct_cleanup`
- `_swift_js_push_pointer`

## Problem

Without `@inline(never)` wrappers, the Swift compiler can inline these public `@_extern(wasm, module: "bjs", ...)` functions across module boundaries. When this happens, the wasm import module attribute changes from `"bjs"` to `"env"` (the default), causing a `wasm-ld` linker error:

```
wasm-ld: error: import module mismatch for symbol: $s13JavaScriptKit17_swift_js_pop_i32s5Int32VyF
>>> defined as bjs in JavaScriptKit.build/BridgeJSIntrinsics.swift.o
>>> defined as env in WASMGoodNotesLibrary.build/BridgeJS.swift.o
```

## Fix

Applied the same pattern already used for f32/f64/pop_pointer: make the `@_extern` declaration `private` and wrap it with a `@inline(never)` public function that prevents cross-module inlining while preserving the correct wasm import module.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/JavaScriptKit/BridgeJSIntrinsics.swift
+++ b/Sources/JavaScriptKit/BridgeJSIntrinsics.swift
@@ -15,8 +15,13 @@ import _CJavaScriptKit
 
 #if arch(wasm32)
 @_extern(wasm, module: "bjs", name: "swift_js_throw")
-@_spi(BridgeJS) public func _swift_js_throw(_ id: Int32)
+private func _swift_js_throw_extern(_ id: Int32)
 #else
+private func _swift_js_throw_extern(_ id: Int32) {
+    _onlyAvailableOnWasm()
+}
+#endif
+
 /// Throws a JavaScript exception from Swift code.
 ///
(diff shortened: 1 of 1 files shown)


kateinoigakukun (approved):

sliemeobn:
@fjtrujy thanks for describing the problem so well here - I was running into the same issue with an ElementaryUI experiment and had no idea what was going on. you saved me hours probably 🙏

PR #633, opened 2026-02-13
NFC: BridgeJS: Reduce retained object cleanup warning noise in BridgeJS glue

krodak:
## Overview

BridgeJS-generated cleanup currently reads retained object IDs with `swift.memory.getObject(id)` before calling `swift.memory.release(id)`. In cases where an ID is already invalid at cleanup time, that pre-read throws first and produces warning noise even though cleanup is already intended to be best-effort.

This change updates `JSGlueGen` to emit a shared retained-object cleanup helper that calls `swift.memory.release(id)` directly inside `try/catch`, and removes the `getObject` pre-check from generated cleanup blocks.


kateinoigakukun (approved):
Thanks!

PR #638, opened 2026-02-16
 BridgeJS: Include pluginGeneratedSources in build plugin input files

krodak:
## Overview

BridgeJS wasn't seeing Swift files generated by other build plugins on the same target. If plugin A generates `@JS`-annotated `.swift` files, BridgeJS (running as plugin B) wouldn't process them because it only looked at `target.sourceFiles`.

Since tools-version 6.0, SPM provides `target.pluginGeneratedSources` - a list of output files declared by previously-run plugins on the same target. Plugin ordering follows declaration order in the `plugins:` array, so this is deterministic.

This change makes BridgeJS include `pluginGeneratedSources` in both its `inputFiles` (for SPM dependency tracking) and the file list passed to BridgeJSTool for skeleton scanning.


kateinoigakukun:
@krodak Thanks for exploring more! `pluginGeneratedSources` approach looks the best for me!

PR #645, opened 2026-02-17
Add AGENTS.md for AI coding agents

krodak:
## Overview

As discussed, here is a starting point for an AGENTS.md file for the JavaScriptKit repo. The goal is to give AI coding agents enough context to contribute without needing to cross-reference CONTRIBUTING.md or reverse-engineer the codebase structure.

This is a draft - let's add what's missing and iterate on it here.

## Notes

The file covers:

- **Project structure and source code layout** - the dual-package architecture with symlinks between `Sources/BridgeJSTool/` and `Plugins/BridgeJS/Sources/`, quick path reference table
- **BridgeJS codegen pipeline** - all three stages (SwiftToSkeleton, ExportSwift/ImportTS, BridgeJSLink), the skeleton IR concept, JS conversion directions
- **Import and export sides** - macro system (`@JS` for exports, `@JSFunction`/`@JSGetter`/`@JSSetter`/`@JSClass` for imports), `CallJSEmission` builder pattern
- **Wasm ABI** - three ABI strategies (value return, stack, side channel), type mapping table, parameter passing conventions, error handling (both directions), closure lifecycle
- **Workflows** - before-committing checklist (from CI config), per-area build/test commands, regeneration steps, adding intrinsics
- **Testing** - snapshot test flow with correct paths (`BridgeJSCodegenTests`, `Inputs/MacroSwift/`), E2E test setup
- **Adding new types** - 10-step checklist from skeleton to docs
- **Common pitfalls** - toolchain, stack ordering, shared snapshots, tmpRetTag overwrite

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,628 @@
+# AGENTS.md - JavaScriptKit
+
+Instructions for AI coding agents contributing to JavaScriptKit.
+
+## Project Overview
+
+JavaScriptKit is a Swift framework for interacting with JavaScript from WebAssembly. The repository contains:
+
+- **Core library** (`Sources/JavaScriptKit/`) - Swift-to-JavaScript interop via `JSObject`, `JSValue`, `JSClosure`, etc.
+- **Companion libraries** (`Sources/JavaScriptEventLoop/`, `Sources/JavaScriptBigIntSupport/`, `Sources/JavaScriptFoundationCompat/`)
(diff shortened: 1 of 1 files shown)


krodak:
Wasn't aware of that one! Makes sense, will push there instead.

Should we add `AGENTS.md` to `.gitignore` in the main repo? That way agents/contributors can fetch it from the skills repo and drop it into their worktree without it showing up as an uncommitted artifact, WDYT?

kateinoigakukun:
@krodak  Sounds good to me :)

PR #656, opened 2026-02-18
NFC: BridgeJS: Refactor JSGlueGen with compositional optional handling and helper consolidation

krodak:
Mix of concrete simplifications and structural refactoring in JSGlueGen.swift. Happy to split this into a smaller PR with just the simplification items if the refactoring part is too much for one review.

### Simplifications

- **Factory collapse**: Struct and enum helper factories collapsed from `() => { return () => ({...}); }` to `() => ({...})` - the double-invocation wrapper was left over from PR #655
- **`simpleEnumHelper`/`rawValueEnumHelper` unified** into `caseEnumHelper` - they had identical logic
- **Dead code removed**: `emitPushI32Return`, `emitPushF64Return`, `emitPushPointerReturn` were copies of their parameter counterparts; several fragment aliases that just forwarded to another fragment
- **`emitRetainCase` helper** extracted to deduplicate 5 identical JSValue retain-and-break cases
- **`jsZeroLiteral` scoped down** from public `WasmCoreType` API in BridgeJSSkeleton to fileprivate in JSGlueGen - nobody else uses it
- **Redundant `(ret | 0)` coercion dropped** from optional case enum returns - the non-optional path never applied it, and the value is already an integer from the enum helper so `| 0` was a no-op

### Structural refactoring

- **Compositional optional handling**: Instead of separate `IntrinsicJSFragment` definitions for every `optional(X)` combination, generic fragments wrap the inner type's fragments. Adding a new `BridgeType` no longer requires writing dedicated optional fragment variants - you set a few properties and optional support composes from the base fragments. The stack ABI optional path is fully generic, which should help with planned external class support.
- **Type-knowledge properties on BridgeType**: `wasmParams`, `optionalConvention`, `nilSentinel`, `isSingleParamScalar`, `lowerCoerce`, etc. moved from free functions and static methods into a `private extension BridgeType` in JSGlueGen.swift. Reads as `type.wasmParams` instead of `wasmParams(for: type)`.
- **`reserveNames` scope change** to fileprivate, moved to bottom of `JSGlueVariableScope`

### Trade-offs

The composite optional wrapper produces slightly more verbose JS than the old per-type fragments (`result` intermediary vars and if/else vs inline ternaries). The codegen is easier to extend and engines optimize both patterns the same way.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
@@ -1219,7 +1219,8 @@ public struct BridgeJSLink {
                 loweringFragment.parameters.count == 1,
                 "Lowering fragment should have exactly one parameter to lower"
             )
-            let loweredValues = try loweringFragment.printCode([param.name], context)
+            let paramName = scope.variable(param.name)
+            let loweredValues = try loweringFragment.printCode([paramName], context)
             parameterForwardings.append(contentsOf: loweredValues)
         }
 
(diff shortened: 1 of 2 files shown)


kateinoigakukun (approved):
Nice!

PR #676, opened 2026-02-25
faster JSObjectSpace (JS runtime retain / release)

sliemeobn:
I added a tiny benchmark to measure the `JSObjectSpace` in isolation and made a few changes to speed things up.

results (two example runs)
```
JSObjectSpaceOriginal    retain: 31.53ms  release: 23.57ms  mixed(1k): 12.32ms  mixed(10k): 15.83ms  mixed(50k): 21.63ms
JSObjectSpace (current)  retain: 10.68ms  release: 14.31ms  mixed(1k): 7.20ms  mixed(10k): 6.37ms  mixed(50k): 9.19ms

JSObjectSpaceOriginal    retain: 27.85ms  release: 24.78ms  mixed(1k): 12.44ms  mixed(10k): 15.76ms  mixed(50k): 22.55ms
JSObjectSpace (current)  retain: 10.96ms  release: 13.62ms  mixed(1k): 6.18ms  mixed(10k): 6.32ms  mixed(50k): 9.63ms
```

I included the benchmark script (AI slop) but I can remove it if we want a clean commit.

Also, please note that there is a behavior difference: refs are not no longer "globally unique" - ie: they will be reused once freed. a "use after free" currently causes an exception - after this PR it could now theoretically go undetected.


kateinoigakukun (approved):
I took a time to review this but it looks good to me!
This change makes the maximum number of JSObjects that can live at the same time limited under `(1 << 22) - 1` (around 4M), but it should be an acceptable constraint practically

krodak:
LGTM, one concern from our side - we have couple of operations that traverse graphs and collect all matching nodes into an array of live JS object references. We're testing with graphs with few millions nodes, which means a single walk could hold millions of concurrent JSObjectSpace references. With SLOT_BITS = 22, the limit is ~4.19M concurrent objects. AFAIU, when you exceed it, `retain()` throws a `RangeError` and the app crashes with no recovery path?

Could we consider bumping SLOT_BITS to 24 (16M objects, 256 generations) or even 26 (64M objects, 64 generations)? The performance characteristics should be identical - it's just changing integer constants for the masks and shifts. The generation counter is a nice safety net but even 64 generations would catch the vast majority of stale ref bugs in practice 🙏🏻 

cc @scottmarchant

sliemeobn:
@krodak thanks for the review, I bumped the bitmask to 24 bits

krodak (approved):

PR #727, opened 2026-04-25
[BridgeJS] Synthesize typed-closure init access from declaration surface (#709)

matthewa26:
## Summary

Fixes #709. The `extension JSTypedClosure where Signature == ... { init(...) }` synthesized by BridgeJS is always emitted as `internal`, so a public `@JSClass` exposing a `JSTypedClosure<...>` parameter cannot be consumed from another target — downstream callers have no way to construct the closure value without hand-rolling a public wrapper.

This change derives the synthesized init's access level from the originating Swift declaration:

- Imported skeleton entries (`ImportedFunctionSkeleton`, `ImportedTypeSkeleton`, `ImportedConstructorSkeleton`, `ImportedGetterSkeleton`, `ImportedSetterSkeleton`) record the source access level (new `BridgeJSAccessLevel` enum: `internal < package < public`, default `internal`).
- `BridgeSkeletonWalker` threads the enclosing decl's access level into `BridgeSkeletonVisitor.visitClosure(...)`. Exported decls reuse the existing `explicitAccessControl: String?` field (`"public"` / `"package"` / `"internal"` / nil → inherit).
- `ClosureSignatureCollectorVisitor` now stores `[ClosureSignature: BridgeJSAccessLevel]` and takes the max access level across every surface that references a given signature, so a closure shape used by both a `public` and an `internal` method becomes `public` (one extension is generated per signature). `signatures: Set<ClosureSignature>` is preserved as a computed view for `BridgeJSLink`.
- `ClosureCodegen.renderClosureHelpers` prefixes the synthesized init with `public` / `package` (or leaves it bare for internal). Mirrors the pattern `JSClassMacro` already uses for `init(unsafelyWrapping:)`.

The user's example from #709 now generates `public init(...)`:

```swift
@JSClass(jsName: "Document") public struct JSDocument {
    @JSFunction public func addEventListener(_ type: String, _ listener: JSTypedClosure<(JSEvent) -> Void>) throws(JSException)
}
// →
extension JSTypedClosure where Signature == (JSEvent) -> Void {
    public init(fileID: StaticString = #fileID, line: UInt32 = #line, _ body: @escaping (JSEvent) -> Void) { ... }
}
```

## Test plan

- [x] New input fixture `SwiftTypedClosureAccess.swift` covering: a public `@JSClass`/`@JSFunction` (→ `public init`), a `package` surface (→ `package init`), an internal-only surface (→ bare `init`), and a closure shape shared between a public and an internal method (→ merges to `public init`).
- [x] `swift test --package-path Plugins/BridgeJS` — all 107 tests in 9 suites pass.
- [x] Codegen JSON snapshots re-recorded to include the new `accessLevel` field on imported decls (defaults to `"internal"` for unchanged fixtures, so no semantic drift).
- [x] `swift build --package-path Examples/Basic` builds clean end-to-end.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
@@ -89,6 +89,38 @@ public enum BridgeContext: Sendable {
     case exportSwift
 }
 
+/// Access level applied to bridge-generated declarations.
+///
+/// Ordering (`Comparable`) reflects visibility breadth, so taking `max` of two
+/// levels yields the more permissive one — used to merge a single closure
+/// signature seen across surfaces with different declared access.
+public enum BridgeJSAccessLevel: String, Codable, Equatable, Hashable, Sendable, Comparable {
+    case `internal`
+    case package
+    case `public`
+
+    public static func < (lhs: BridgeJSAccessLevel, rhs: BridgeJSAccessLevel) -> Bool {
+        lhs.order < rhs.order
+    }
+
+    private var order: Int {
+        switch self {
+        case .internal: return 0
+        case .package: return 1
+        case .public: return 2
+        }
+    }
+
+    /// Returns the modifier keyword to emit, or nil for the default (internal).
+    public var modifierKeyword: String? {
+        switch self {
+        case .internal: return nil
+        case .package: return "package"
+        case .public: return "public"
+        }
+    }
+}
+
 public struct ClosureSignature: Codable, Equatable, Hashable, Sendable {
     public let parameters: [BridgeType]
     public let returnType: BridgeType
@@ -398,17 +430,34 @@ public struct Parameter: Codable, Equatable, Sendable {
 // MARK: - BridgeSkeleton Visitor
 
 public protocol BridgeSkeletonVisitor {
-    mutating func visitClosure(_ signature: ClosureSignature, useJSTypedClosure: Bool)
+    /// Called when a closure type is encountered during traversal.
+    ///
+    /// `accessLevel` reflects the source access of the enclosing declaration,
+    /// so visitors can derive an appropriate access level for any
+    /// bridge-generated helpers tied to this signature (e.g. typed closure
+    /// inits in `ClosureCodegen`).
+    mutating func visitClosure(
+        _ signature: ClosureSignature,
+        useJSTypedClosure: Bool,
+        accessLevel: BridgeJSAccessLevel
+    )
     mutating func visitImportedFunction(_ function: ImportedFunctionSkeleton)
 }
 
 public extension BridgeSkeletonVisitor {
-    mutating func visitClosure(_ signature: ClosureSignature, useJSTypedClosure: Bool) {}
+    mutating func visitClosure(
+        _ signature: ClosureSignature,
+        useJSTypedClosure: Bool,
+        accessLevel: BridgeJSAccessLevel
+    ) {}
     mutating func visitImportedFunction(_ function: ImportedFunctionSkeleton) {}
 }
 
 public struct BridgeSkeletonWalker<Visitor: BridgeSkeletonVisitor> {
     public var visitor: Visitor
+    /// Tracks the access level of the enclosing declaration during traversal.
+    /// Saved/restored around each declaration that introduces an access boundary.
+    private var currentAccessLevel: BridgeJSAccessLevel = .internal
 
     public init(visitor: Visitor) {
         self.visitor = visitor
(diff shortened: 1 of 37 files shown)


Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -1180,16 +1289,32 @@ public struct ImportedModuleSkeleton: Codable {
...
 
-    public mutating func visitClosure(_ signature: ClosureSignature, useJSTypedClosure: Bool) {
-        signatures.insert(signature)
+    public mutating func visitClosure(
+        _ signature: ClosureSignature,
+        useJSTypedClosure: Bool,
+        accessLevel: BridgeJSAccessLevel
+    ) {
+        if let existing = signatureAccessLevels[signature] {
+            signatureAccessLevels[signature] = max(existing, accessLevel)
```
krodak, line 1415: The "check existing, take max, else insert" pattern here is duplicated in `recordInjectedSignature` below. If the merge logic ever needs to change (e.g. adding a diagnostic for conflicting levels), you'd need to update both spots.

Small extract:

```swift
private mutating func recordSignature(
    _ signature: ClosureSignature,
    accessLevel: BridgeJSAccessLevel
) {
    if let existing = signatureAccessLevels[signature] {
        signatureAccessLevels[signature] = max(existing, accessLevel)
    } else {
        signatureAccessLevels[signature] = accessLevel
    }
}
```

Then both `visitClosure` and `recordInjectedSignature` call through to it.

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -532,6 +595,30 @@ public struct BridgeSkeletonWalker<Visitor: BridgeSkeletonVisitor> {
...
+    }
+
+    /// String-typed convenience: maps `"public"`/`"package"`/`"internal"` from
+    /// `Exported*.explicitAccessControl` to the typed enum.
+    private mutating func withAccessLevel(
+        _ rawLevel: String?,
+        _ body: (inout BridgeSkeletonWalker) -> Void
+    ) {
+        withAccessLevel(rawLevel.flatMap(BridgeJSAccessLevel.init(rawValue:)), body)
+    }
```
krodak, line 631: `rawLevel.flatMap(BridgeJSAccessLevel.init(rawValue:))` silently drops unknown strings (e.g. `"open"`, `"private"`) and falls back to inheriting the outer level. That's fine today since the macros reject those, but it's a quiet invariant. An assertion for unexpected values would save debugging time if the exported side ever gains new access strings:

```swift
private mutating func withAccessLevel(
    _ rawLevel: String?,
    _ body: (inout BridgeSkeletonWalker) -> Void
) {
    let level: BridgeJSAccessLevel?
    if let rawLevel {
        level = BridgeJSAccessLevel(rawValue: rawLevel)
        assert(level != nil, "Unexpected access level string: \(rawLevel)")
    } else {
        level = nil
    }
    withAccessLevel(level, body)
}
```

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -1180,16 +1289,32 @@ public struct ImportedModuleSkeleton: Codable {
...
+    /// set of unique signatures, without access metadata.
+    public var signatures: Set<ClosureSignature> { Set(signatureAccessLevels.keys) }
     let moduleName: String
 
     public init(moduleName: String, signatures: Set<ClosureSignature> = []) {
         self.moduleName = moduleName
-        self.signatures = signatures
+        for signature in signatures {
+            signatureAccessLevels[signature] = .internal
+        }
```
krodak, line 1394: This seeds every pre-existing signature as `.internal`. That's correct for the only current caller (`BridgeJSLink`, exported side), but the API doesn't communicate the assumption. If someone later pre-seeds signatures that should be public, they'd silently get capped.

Two options (both low-effort):

1. Add a doc comment on this init noting the assumption:
```swift
/// Convenience for callers that only need to seed signatures without
/// access metadata (e.g. exported-side walking where closure init
/// access is irrelevant). All seeded signatures default to `.internal`.
public init(moduleName: String, signatures: Set<ClosureSignature>) {
```

2. Or offer a dictionary-based init alongside it:
```swift
public init(moduleName: String, signatureAccessLevels: [ClosureSignature: BridgeJSAccessLevel] = [:]) {
    self.moduleName = moduleName
    self.signatureAccessLevels = signatureAccessLevels
}
```

krodak (approved):
Hey Matthew, appreciate clean fix for a real usability problem 👌🏻 
The approach of threading access levels through the skeleton/walker and merging via `max` fits the existing architecture well. Test fixture covers the key scenarios. A few suggestions below, nothing blocking.

krodak:
@matthewa26 please address feedback at your convenience and have a read on `CONTRIBUTING.md` to make sure to update artifacts checked by CI 🙏🏻

matthewa26:
@krodak Will do! Just now seeing this.

matthewa26:
@krodak Looks like everything is green now. Thanks for the review and the follow up!

kateinoigakukun (approved):
Seems good to me, thanks @matthewa26 (and thanks @krodak for reviewing!)

Issue #720, opened 2026-04-14
[BridgeJS] Support `@JS` type references across modules within the same package


wfltaylor:
Larger packages are often split into multiple modules. For example, a package could contain a foundational `Vector3D` type in a `Core` module which is used by the public interface of the `App` module.

```swift
// “Core” Module

@JS public struct Vector3D {

    let x: Double
    let y: Double
    let z: Double

}

// “App” Module

import Core

@JS func currentVelocity() -> Vector3D {
    // ...
}
```

Currently, the “App” module would have to wrap any types defined in other modules, or everything would have to be defined in the one module. Both of these workarounds are undesirable.

This could possibly work by parsing the `BridgeJS.json` skeleton of dependent targets in the code generation plugin. Further thought would probably be needed regarding ahead of time generation.

krodak:
Tested on [0.51.0](https://github.com/swiftwasm/JavaScriptKit/releases/tag/0.51.0), works well 🎉

PR #730, opened 2026-04-28
BridgeJS: Use `@JS` types from other modules in the same package

wfltaylor:
Support using `@JS` types from one internal module in `@JS` API in another.

This works by parsing the BridgeJS skeleton of each dependency, and using that to resolve types if we fail to find a match within the target. The skeleton of the dependency is passed to the tool using `--dependency-skeleton <Module>=<path>`. For this to work with the `BridgeJS` build plugin, the dependency must contain a BridgeJS config file to declare that it uses BridgeJS.

There is no attempt to match Swift’s type resolution exactly: most ambiguity requires the module to be explicitly specified. Local types will shadow external types matching Swift’s behaviour.

This doesn’t support [module selectors](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0491-module-selectors.md), since I think this would require bumping the minimum SwiftSyntax version.

This doesn’t support using types from modules in other packages. Extending an `@JS` type defined in another module is also unsupported.

A required field is introduced into the skeleton format. As far as I am aware this shouldn’t be an issue, since there is no need for the skeleton format to be backwards compatible at this time. If there is in fact such need, it should be trivial to make this optional when decoding.

Addresses issue https://github.com/swiftwasm/JavaScriptKit/issues/720


kateinoigakukun (commented):
Thanks for working on this! I still haven't taken a closer look at the diff but a few questions:

1. Do we support AOT codegen cases? To codegen AOT, we need to require all dependency modules to AOT codegen as well to get skeletons. Can we produce a clear diagnostic when BridgeJS plugin is configured for a dependency module  but not AOTed.
2. Did you see any particular blocker for supporting cross-package reference?

wfltaylor:
> Thanks for working on this! I still haven't taken a closer look at the diff but a few questions:
> 
> 1. Do we support AOT codegen cases? To codegen AOT, we need to require all dependency modules to AOT codegen as well to get skeletons. Can we produce a clear diagnostic when BridgeJS plugin is configured for a dependency module  but not AOTed.

Yep, this is supported. I’ve pushed a commit to emit a diagnostic when attempting to AOT codegen a target which depends on a BridgeJS enabled target which has not been AOT codegened. If you don’t explicitly specify the target, it will AOT codegen them all in the correct order.

> 2. Did you see any particular blocker for supporting cross-package reference?

I tried to keep the scope of this PR smaller so I didn’t look into it too much, but this should be possible as long as we can find the skeleton file. I’m not sure how build plugins work in package dependencies, but I believe the worst case would be requiring the dependency to use AOT codegen for this use case.

wfltaylor:
> > Thanks for working on this! I still haven't taken a closer look at the diff but a few questions:
> > 
> > 1. Do we support AOT codegen cases? To codegen AOT, we need to require all dependency modules to AOT codegen as well to get skeletons. Can we produce a clear diagnostic when BridgeJS plugin is configured for a dependency module  but not AOTed.
> 
> Yep, this is supported. I’ve pushed a commit to emit a diagnostic when attempting to AOT codegen a target which depends on a BridgeJS enabled target which has not been AOT codegened. If you don’t explicitly specify the target, it will AOT codegen them all in the correct order.
> 
> > 2. Did you see any particular blocker for supporting cross-package reference?
> 
> I tried to keep the scope of this PR smaller so I didn’t look into it too much, but this should be possible as long as we can find the skeleton file. I’m not sure how build plugins work in package dependencies, but I believe the worst case would be requiring the dependency to use AOT codegen for this use case.

Or I guess we could just re-parse the source assuming we are able to locate that.

Plugins/BridgeJS/Tests/BridgeJSToolTests/CrossModuleResolutionTests.swift
```diff
@@ -0,0 +1,491 @@
...
+            )
+            Issue.record("Expected unsupported-type diagnostic for Foundation.URL")
+        } catch let error as BridgeJSCoreDiagnosticError {
+            let combinedMessages = error.diagnostics.map(\.diagnostic.message).joined(separator: "\n")
+            #expect(combinedMessages.contains("Foundation.URL"))
+        }
+    }
+
+    // MARK: - Utillities
+
```
krodak, line 461: Nit: `Utillities` → `Utilities` (double-L typo)

krodak (approved):
Hey William,

Really like how this came together. The `ExternalModuleIndex` is a clean abstraction, the disambiguation via module qualifiers follows Swift conventions naturally, and local-shadows-external is the right default. The dependency ordering in the command plugin and the build plugin skeleton discovery are well thought through.

Thanks for the detailed comment in `BridgeJSTool.swift` about the skeleton-to-Swift injective mapping invariant and the hash escape hatch if it ever breaks. That kind of forward-looking note saves a lot of head-scratching later.

Test coverage is solid across type categories, structural positions, multi-module scenarios, and error diagnostics. The AOT diagnostic from commit 2 and the resource-bundling fix in commit 3 are both good catches.

LGTM!

Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Unsupported-Features.md
```diff
@@ -8,13 +8,11 @@ BridgeJS generates glue code per Swift target (module). Some patterns that are v
 
 ## Type usage crossing module boundary
 
-BridgeJS does **not** support using a type across module boundaries in the following situations.
+### Exporting Swift: extending types from another Swift module
```
kateinoigakukun, line 11: Also we need to note that we cannot use non-`@JS` types defined in external modules. We can relax this requirement by falling back to stack ABI for such foreign unknown types though.

kateinoigakukun (approved):
Overall looks good to me!

PR #738, opened 2026-05-01
BridgeJS: Move optional JSObject to stack ABI, enabling Optional<@JSClass>

krodak:
## Overview

Fixes #666. Moves `Optional<JSObject>` (and by extension `Optional<@JSClass>`) from the side-channel ABI to the stack ABI, enabling optional custom types on both import and export sides.

Previously, `@JSFunction func foo() -> Foo?` where `Foo` is a `@JSClass` struct would fail because the side-channel path for `JSObject` optionals lacked a `bridgeJSLiftReturnFromSideChannel` implementation. Rather than adding another per-type side-channel intrinsic, this switches `jsObject` optionals to the stack ABI — the same convention already used by structs, arrays, and dictionaries.

```swift
@JSClass struct Foo {
    @JSGetter var id: String
}
@JSFunction func foo() throws(JSException) -> Foo?  // now works
```

**What changed:**

**1. `JSGlueGen.swift`** — `optionalConvention` for `.jsObject`: changed from `.sideChannelReturn(.retainedObject)` to `.stackABI`. Removed `jsObject` from `nilSentinel` (stack ABI uses the isSome flag instead).

**2. `BridgeJSSkeleton.swift`** — `usesSideChannelForOptionalReturn()`: moved `.jsObject` from the `true` branch to `false`.

**3. `ImportTS.swift`** — `liftingReturnInfo` for `.nullable`: when the wrapped type no longer uses side-channel, returns `valueToLift: nil` so the extern returns void (values go through typed stacks).

**4. `BridgeJSIntrinsics.swift`** — `Optional<JSObject>.bridgeJSLowerReturn()`: changed from calling `_swift_js_return_optional_object` (global side-channel) to `Wrapped.bridgeJSStackPushAsOptional` (stack ABI). JSObject already conforms to `_BridgedSwiftStackType`, and `_JSBridgedClass` inherits from it, so `Optional<Foo>` gets stack-based lifting/lowering for free through the protocol chain.

**Tests:** Added `childOrNull` getter/setter/roundTrip for `WithOptionalJSClass?` to the `Optionals` test fixture, covering the import-side `@JSClass` optional pattern from the original issue.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
@@ -128,7 +128,9 @@ public struct ImportTS {
             self.returnType = returnType
             self.context = context
             let liftingInfo = try returnType.liftingReturnInfo(context: context)
-            if effects.isAsync || returnType == .void || returnType.usesSideChannelForOptionalReturn() {
+            if effects.isAsync || returnType == .void || returnType.usesSideChannelForOptionalReturn()
+                || liftingInfo.valueToLift == nil
+            {
                 abiReturnType = nil
             } else {
                 abiReturnType = liftingInfo.valueToLift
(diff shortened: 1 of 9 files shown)


krodak:
@kateinoigakukun let me know if this aligns with your comment of revisiting side-channel ABI, I remembered that we've discussed moving this to stack ABI, but might be wrong 😅  🙏🏻

kateinoigakukun (approved):

PR #739, opened 2026-05-04
BridgeJS: Fix closures with struct return

wfltaylor:
Currently, the following code will result in generated code that fails to compile.
```swift
@JS public struct Animal {
    public let type: String
}

@JS func roundtripAnimal(_ animalClosure: (Animal) -> Animal) -> (Animal) -> Animal
```

The generated code uses `bridgeJSLiftReturn` with no arguments, which doesn’t exist.

This fixes this by adding an overload for `bridgeJSLiftReturn` which uses the stack ABI.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
@@ -1441,8 +1441,6 @@ public struct BridgeJSLink {
                 }
             }
             return type.tsType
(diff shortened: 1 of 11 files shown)


wfltaylor:
@krodak @kateinoigakukun I’ve pushed a semi-related fix which the new tests here surfaced in the generation of the TypeScript type definitions, where `Tag` was incorrectly emitted as a suffix to the type. Let me know if you want me to split this into a separate PR.

krodak (approved):
All looks good 👌🏻

PR #742, opened 2026-05-10
fix: use `export type` for SwiftRuntimeThreadChannel re-export

### LINKED ISSUE #477: JavaScript runtime portability

shivanshu877:
Fixes https://github.com/swiftwasm/JavaScriptKit/issues/477

`SwiftRuntimeThreadChannel` is a TypeScript type alias, not a value. Re-exporting it with the value-style `export { ... }` causes esbuild and rolldown (which run in `isolatedModules` mode by default) to fail with:

```
Cannot re-export a type when 'isolatedModules' is enabled.
```

This breaks anyone trying to bundle JavaScriptKit with Vite or other esbuild/rolldown-based toolchains.

Changing to `export type { SwiftRuntimeThreadChannel }` makes the intent explicit and restores compatibility with esbuild/rolldown without affecting `tsc` or any runtime behaviour — a value-only consumer wouldn't be importing this anyway.

## Test plan

- [x] `tsc` still compiles cleanly.
- [x] Type-only consumers still resolve the import.
- [ ] Local verification with esbuild / rolldown is welcome from a maintainer with the toolchain set up.

### DIFF
--- a/Runtime/src/index.ts
+++ b/Runtime/src/index.ts
@@ -20,7 +20,7 @@ import {
 } from "./itc.js";
 import { decodeObjectRefs } from "./js-value.js";
 import { JSObjectSpace } from "./object-heap.js";
-export { SwiftRuntimeThreadChannel };
+export type { SwiftRuntimeThreadChannel };
 
 export type SwiftRuntimeOptions = {
     /**


shivanshu877:
Local verification (per @krodak's earlier request):

Ran on macOS 26 / Apple Swift 6.3.2:
- `cd Runtime && npm install && npm run build` (Rollup) — **succeeds** with the `export type { SwiftRuntimeThreadChannel }` change. The whole point of the fix is making `isolatedModules`-mode bundlers (esbuild, rolldown) happy; the build going through rollup cleanly is the closest local equivalent.
- `swift test --package-path Plugins/BridgeJS` — **all 134 tests in 10 suites pass** (BridgeJSCodegenTests, BridgeJSLinkTests, JSClassMacroTests, etc.). No regressions from the re-export change.

(JavaScriptKit's full test suite needs the SwiftWasm toolchain + carton to execute the WebAssembly-side tests; those require the project's CI runners. The Mac-native subset above is what's directly affected by this PR.)

krodak:
@shivanshu877 

> (JavaScriptKit's full test suite needs the SwiftWasm toolchain + carton to execute the WebAssembly-side tests; those require the project's CI runners. The Mac-native subset above is what's directly affected by this PR.)

Carton is not required, you can test locally with SwiftWasm toolchain, you will find instructions in .md files in repo. you might find those skills useful as well: https://github.com/swiftwasm/Swift-Wasm-Agent-Skill

PR #744, opened 2026-05-11
BridgeJS: Fix name collision for same-named nested structs

krodak:
## Overview

Follow-up to #735. When two different parent types both have a nested struct with the same short name (e.g., `User.Stats` and `Player.Stats`), the generated JavaScript produces duplicate `const` declarations and overwrites the same `structHelpers.Stats` key, causing a SyntaxError at parse time.

**1. JS struct helpers use `abiName` (underscore-separated) for unique internal keys.** `structHelpers.User_Stats` and `structHelpers.Player_Stats` instead of both being `structHelpers.Stats`. This covers helper factories, registration, and all `structHelpers.<name>.lower/lift()` call sites in both `BridgeJSLink.swift` and `JSGlueGen.swift`.

**2. TypeScript declarations use TS namespaces for nested struct interfaces.** Instead of `export interface User_Stats`, the `.d.ts` now emits:

```typescript
export namespace User {
    export interface Stats { ... }
}
```

This matches the existing namespace pattern used for enums and keeps the public API hierarchical. The `NamespaceBuilder` was extended to handle structs alongside enums/classes. Struct-only namespace nodes are excluded from the `Exports` type to avoid duplicate keys.

**3. `tsType` for `.swiftStruct` returns the dotted name directly** (e.g., `User.Stats`), which TypeScript resolves through the namespace. Non-nested structs are unaffected.

**4. E2e runtime test added.** `NestedStructGroupA.Metadata` and `NestedStructGroupB.Metadata` (two same-named structs under different namespace enums) are roundtripped through JS to verify no data corruption from helper key collision.

### DIFF
--- a/Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/NestedType.d.ts
+++ b/Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/NestedType.d.ts
@@ -4,9 +4,17 @@
 // To update this file, just rebuild your project or run
 // `swift package bridge-js`.
 
-export interface Stats {
-    health: number;
-    score: number;
+export namespace Player {
+    export interface Stats {
+        level: number;
+        rating: string;
+    }
+}
+export namespace User {
+    export interface Stats {
+        health: number;
+        score: number;
+    }
 }
 /// Represents a Swift heap object like a class instance or an actor instance.
 export interface SwiftHeapObject {
@@ -18,8 +26,13 @@ export interface SwiftHeapObject {
 export interface User extends SwiftHeapObject {
     getName(): string;
 }
+export interface Player extends SwiftHeapObject {
+    getTag(): string;
+}
 export type Exports = {
     User: {
+    }
+    Player: {
     }
 }
 export type Imports = {
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
@@ -232,7 +232,9 @@ public struct BridgeJSLink {
             var structExportEntries: [(js: [String], dts: [String])] = []
             for structDefinition in skeleton.structs {
                 let (jsStruct, dtsType, dtsExportEntry) = try renderExportedStruct(structDefinition)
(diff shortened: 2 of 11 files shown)


Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/NestedType.d.ts
```diff
@@ -4,10 +4,14 @@
 // To update this file, just rebuild your project or run
 // `swift package bridge-js`.
 
-export interface Stats {
+export interface User_Stats {
```
wfltaylor: I wonder if `UserStats` would be more aesthetically pleasing? I guess this could conflict with a separate struct just called `UserStats` in Swift.

More work but I think we could use TS namespaces here as well:
```
export namespace User {
    export interface Stats {
    }
}
```
I believe that this will automatically merge with the `interface Player`, if it exists.

krodak: Hey William, on the naming - yeah, `UserStats` would read nicer, but as you noted it could collide with an actual `UserStats` struct. The underscore keeps it unambiguous and mirrors the internal ABI name and aligns with current approach elsewhere in the code.

As for TS namespace idea - we do already emit namespace blocks in the declare global path (for `exposeToGlobal: true`), so the mechanism exists and could be reused. Current approach aligns with use of underscore in other places, but that could be considered, depending on @kateinoigakukun preference.

kateinoigakukun: I think it makes more sense to use TS-namespace to represent the hierarchical types for public user-facing APIs.

krodak: Updated - nested struct interfaces now use TS namespaces. `User.Stats` and `Player.Stats` generate:

```typescript
export namespace User {
    export interface Stats {
        health: number;
        score: number;
    }
}
export namespace Player {
    export interface Stats {
        level: number;
        rating: string;
    }
}
```

Internal JS struct helpers still use underscore-separated keys (`structHelpers.User_Stats`) since those aren't user-facing.

kateinoigakukun (approved):

PR #745, opened 2026-05-13
BridgeJS: Optimize numeric array transfer with bulk TypedArray copy

krodak:
## Overview

Optimize numeric array transfer by using bulk TypedArray memory copy instead of element-by-element stack serialization. When a Swift function passes or receives a numeric array (`[Int]`, `[UInt8]`, `[Float]`, `[Double]`, etc.), BridgeJS now transfers the data as a single bulk operation. TypeScript types remain `number[]` / `bigint[]` — no user-facing API change.

This is a transparent performance optimization following the same pattern as `bridgeJSStackPopAsOptional` — a specialized ABI for numeric element types without changing the type contract. Non-numeric arrays (`[String]`, `[MyStruct]`, etc.) continue using the existing element-by-element stack protocol.

**1. Swift → JS direction:** `Array.bridgeJSTypedArrayPush()` calls `withUnsafeBufferPointer` to pass `(ptr, count, kind)` to a new `swift_js_push_typed_array` WASM import, which copies the bytes into a JS TypedArray. A pre-allocated for-loop then converts to `number[]` for the caller.

**2. JS → Swift direction:** `[T].bridgeJSTypedArrayLiftParameter(sourceId, count)` receives a retained TypedArray from JS, allocates via `Array(unsafeUninitializedCapacity:)`, and calls back to JS via `swift_js_init_typed_array_memory` to bulk copy — the same retain-allocate-callback pattern used by string lowering.

**3. Affected types:** All integer types (Int8 through Int64, signed and unsigned), Float, Double, plus pointer-width Int/UInt. A new `_BridgeJSTypedArrayElement` protocol marks eligible types. `BridgeType.isNumericScalar` and `typedArrayKind` on the skeleton side route the codegen.

### Benchmark results

Release build, 10K iterations × 5 runs:

| Test | Before | After | Speedup |
|------|--------|-------|---------|
| takeIntArray (1K, JS→Swift) | 758 ms | 13 ms | **58×** |
| takeDoubleArray (1K, JS→Swift) | 791 ms | 13 ms | **61×** |
| makeIntArray (1K, Swift→JS) | 221 ms | 18 ms | **12×** |
| makeIntArrayLarge (10K, Swift→JS) | 2160 ms | 146 ms | **15×** |
| roundtripIntArray (1K) | 964 ms | 29 ms | **33×** |
| takeStringArray (non-numeric) | 23 ms | 25 ms | unchanged |

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
@@ -347,6 +347,7 @@ public struct BridgeJSLink {
             "let \(JSGlueVariableScope.reservedF32Stack) = [];",
             "let \(JSGlueVariableScope.reservedF64Stack) = [];",
             "let \(JSGlueVariableScope.reservedPointerStack) = [];",
+            "let \(JSGlueVariableScope.reservedTaStack) = [];",
             "const \(JSGlueVariableScope.reservedEnumHelpers) = {};",
             "const \(JSGlueVariableScope.reservedStructHelpers) = {};",
             "",
@@ -487,6 +488,21 @@ public struct BridgeJSLink {
                 printer.write("bjs[\"swift_js_pop_i64\"] = function() {")
                 printer.indent {
                     printer.write("return \(JSGlueVariableScope.reservedI64Stack).pop();")
+                }
+                printer.write("}")
+                // Typed array constructors indexed by kind (must match _BridgedNumericArrayKind)
+                printer.write(
+                    "const taCtors = [Int8Array, Uint8Array, Int16Array, Uint16Array, Int32Array, Uint32Array, Float32Array, Float64Array];"
+                )
+                printer.write("bjs[\"swift_js_push_typed_array\"] = function(kind, ptr, count) {")
+                printer.indent {
+                    printer.write("const Ctor = taCtors[kind];")
+                    printer.write("const byteLen = count * Ctor.BYTES_PER_ELEMENT;")
+                    // slice() copies the bytes into a new ArrayBuffer that is properly aligned
+                    printer.write(
+                        "const copy = \(JSGlueVariableScope.reservedMemory).buffer.slice(ptr, ptr + byteLen);"
+                    )
+                    printer.write("\(JSGlueVariableScope.reservedTaStack).push(Array.from(new Ctor(copy)));")
                 }
                 printer.write("}")
                 if !allStructs.isEmpty {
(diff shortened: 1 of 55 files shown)


Plugins/BridgeJS/Sources/BridgeJSCore/ClosureCodegen.swift
```diff
@@ -132,6 +132,19 @@ public struct ClosureCodegen {
 
         for (index, paramType) in signature.parameters.enumerated() {
             let paramName = "param\(index)"
+
+            // Numeric arrays use bulk TypedArray transfer with (sourceId, count) WASM params
+            if case .array(let elementType) = paramType, elementType.isNumericScalar {
```
kateinoigakukun: This special casing is what I'd like to avoid if possible. 

I think we can remove most of the codegen changes around call emission if we do something like:

```swift
protocol _BridgedSwiftStackType {
    /// Specialization point for popping an `Array<Self>`
    static func bridgeJSStackPopAsArray() -> Array<StackLiftResult>
    /// Specialization point for pushing an `Array<Self>`
    static func bridgeJSStackPushAsArray(_ value: consuming Array<Self>)
}

extension _BridgedSwiftStackType {
    static func bridgeJSStackPopAsArray() -> Array<StackLiftResult> {
        // Bring the default generic pop impl
        // https://github.com/swiftwasm/JavaScriptKit/blob/785e1e2daad01c8bedcff76a47ff855bc1e197b5/Sources/JavaScriptKit/BridgeJSIntrinsics.swift#L2108
    }
    static func bridgeJSStackPushAsArray(_ value: consuming Array<Self>) { /* the same as pop */ }
}

extension Array: _BridgedSwiftStackType where Element: _BridgedSwiftStackType, Element.StackLiftResult == Element {
    @_spi(BridgeJS) public static func bridgeJSStackPop() -> [Element] {
        // CHANGE: Just to call bridgeJSStackPushAsArray
        Element.bridgeJSStackPushAsArray(self)
    }
    @_spi(BridgeJS) public consuming func bridgeJSStackPush() { /* the same as pop */ }
}

extension UInt8: _BridgedSwiftTypeLoweredIntoSingleWasmCoreType, _BridgedSwiftStackType {
    static func bridgeJSStackPopAsArray() -> Array<UInt8> {
        self.withUnsafeBufferPointer { buffer in /* blah blah */ }
    }
}
```

I believe we can use the conventional lift/lower methods without recognizing specializable types in codegen.

kateinoigakukun (approved):

Issue #737, opened 2026-05-01
Support exporting structs as a reference


wfltaylor:
The current mechanism for exporting structs works great for simple structs, like `@JS struct Point { let x: Double; let y: Double }`. When passed out or into Swift, all the members of the struct are copied.

However, it is also common in Swift to have a less trivial struct, for example:
```
@JS struct Polygon {

    var vertices: [Point]

}
```
A complex polygon could have tens of thousands of vertices. This works fine in Swift, since the Array is COW. However, this is less than ideal when interoperating with JavaScript, where all the vertices would have to be copied every time the struct crosses the boundary.

[Mutating functions currently don’t work on structs](https://github.com/swiftwasm/JavaScriptKit/issues/736), but if they were working, this would also require copying the whole contents of the struct after each call. For example, a mutating `addVertex` function on `Polygon` would require copying every vertex across the boundary.

To workaround this issue, you could convert the struct into a class. This improves interoperability with JavaScript, however it can make Swift code less idiomatic.

### Proposed Solution

A solution would be allowing structs to be exported using a heap allocated box, which could be expressed using `@JS(boxed: true)`. A “boxed” struct would lead to very similar generated code to a class, where JavaScript just has a Swift-owned pointer.

krodak:
Hey @wfltaylor, I understand below suggestion does not resolve the issue, but it might help until we have boxing mechanism in place.

If you switch both `Polygon` and `Point` to classes and enable [identity mode](https://swiftpackageindex.com/swiftwasm/javascriptkit/0.51.0/documentation/javascriptkit/exporting-swift-class#Identity-Mode), BridgeJS maintains a `WeakRef`-based cache keyed by the Swift heap pointer. The same Swift object always returns the same JS wrapper, so you avoid recreating wrapper objects on every boundary crossing. `Point` needs to be a class too, otherwise the `[Point]` array still copies every vertex as a value type when it crosses the boundary.

Per-class opt-in with the macro:

```swift
@JS(identityMode: true)
class Point {
    @JS var x: Double
    @JS var y: Double
    @JS init(x: Double, y: Double) { self.x = x; self.y = y }
}

@JS(identityMode: true)
class Polygon {
    @JS var vertices: [Point]
    // ...
}
```
There's also a [project-wide option](https://swiftpackageindex.com/swiftwasm/javascriptkit/0.51.0/documentation/javascriptkit/bridgejs-configuration#identityMode) in bridge-js.config.json:

```json
[...]

Issue #496, opened 2026-01-15
[BridgeJS] Support Foundation.URL as parameter/return type


kateinoigakukun:

MaxDesiatov:
However this is implemented, it should be configurable for users building with Embedded Swift, where Foundation is unavailable.

krodak:
I was thinking about working on implementing some Foundation types support in BridgeJS, starting with `UUID` (which could simplify interface a bit in our project) and designing the mechanism that could also cover `URL`

**Proposed approach:**

Add a new `BridgeType.foundationType(FoundationTypeKind)` case where `FoundationTypeKind` is a closed enum of recognized Foundation types (`uuid`, `url`, extensible to `date` etc.). Each kind declares its wire type (e.g., both UUID and URL use `string` on the wire), Swift type name, and TypeScript type.

This means:
- **ABI:** Foundation types reuse existing wire encodings (UUID/URL → string)
- **Swift side:** Auto-generated lift/lower code converts between the Foundation type and its wire representation (`UUID(uuidString:)!` / `.uuidString`, `URL(string:)!` / `.absoluteString`)
- **TypeScript side:** Maps to `string`
- **Embedded Swift:** All Foundation bridging gated behind `#if canImport(Foundation)` in `BridgeJSIntrinsics.swift`, following approach similar to `JavaScriptFoundationCompat`

The single `foundationType` case keeps changeset manageable,  most switch statements over `BridgeType` can delegate to the wire type's existing handling rather than needing per-type branches.

I could start on this early next week, happy to discuss the design before I start coding 🙏🏻

kateinoigakukun:
Thanks @krodak for sharing your idea

TBH I'm a bit concerned about adding more BridgeType cases before we pay down some of the existing complexity debt in the code generator. In particular, the current ad-hoc handling for container types should probably be moved out of the code generator and into something that leverages Swift's generic type system more directly.

I think we should limit the set of types that the code generator needs to recognize to only those that truly require specialized, performance-critical handling. For everything else, I've been wondering whether we could adopt a more generic ABI approach (similar in spirit to a stack-based ABI), so that extending supported types wouldn't require invasive changes to the code generator.

PR #750, opened 2026-05-27
BridgeJS: Export types using a separate JS representation

wfltaylor:
Note that this PR is much smaller than it looks! The vast majority of the lines added are tests/generated code.

Adds `@JS(as:)` to let types provide an alternative JavaScript representation, providing an escape hatch when BridgeJS’s defaults don't fit.

# Motivation

BridgeJS provides sensible defaults when exporting Swift code to JavaScript. For example, structs are exported using copy semantics: each field in the Swift struct is copied when crossing the boundary. BridgeJS also only supports exporting a subset of Swift features, which makes sense given there are many things in the Swift language which would be difficult to expose to JavaScript.

While this approach makes sense, sometimes the defaults used by BridgeJS won’t work well in a specific situation. For example, a Swift `struct Polygon { var vertices: [Point] }` will be exported to JavaScript using copy semantics. In a project where polygons could contain thousands of points, copying at the boundary every time provides unacceptable performance. Currently, there is no escape hatch for situations like these. Either the Swift code has to be modified to be less idiomatic (e.g. in this case by making `Polygon` a class), or a duplicate type could be created (say a `JSPolygon`) and exposed to JavaScript that wraps the underlying Swift type. A similar situation could be encountered when using a feature which BridgeJS doesn’t support, for example dictionaries with integer keys. The same two options are available: less idiomatic Swift code (using String keys everywhere), or duplicating types.

At first, the “duplicated types” approach sounds reasonable. Unfortunately, it starts to break down as the project scales. For example, a large existing codebase could use `Polygon` **in hundreds of types and hundreds of methods**. Creating a duplicate JS hierarchy here introduces a huge maintenance cost, where most of the duplication is providing no value since BridgeJS’s defaults likely make sense the majority of the time. The same situation occurs with unsupported features: a leaf type might introduce a dictionary with integer keys, requiring the creation of a vast hierarchy of duplicated JS-safe types.

# Solution

Allow types to provide an alternative JS representation. This works as follows:

```swift
@JS(as: JSPolygon.self) public struct Polygon {

    public var vertices: [Point]

    public consuming func bridgeToJS() -> JSPolygon {
        JSPolygon(underlying: self)
    }

    public static func bridgeFromJS(_ value: consuming JSPolygon) -> Polygon {
        value.underlying
    }

}

@JS public final class JSPolygon {

    var underlying: Polygon

    @JS public init(underlying: Polygon) {
        self.underlying = underlying
    }

    @JS var boundingBox: Rect { underlying.boundingBox }

}
```

Now, the existing Swift code can continue to use the Swift-idiomatic `Polygon` while JavaScript gets the appropriate `JSPolygon` type.

# Design

This is implemented by inserting calls to `bridgeToJS()` and `bridgeFromJS` at the first and last possible opportunity respectively. This means that code generation changes are minimal, everything just uses the ABI of the JavaScript type. I’ve added a new case to `BridgeType` `indirect case alias(name: String, underlying: BridgeType)`. I’m very much not sold on the name `alias`, but I couldn’t think of anything better.

# Alternatives

- PR https://github.com/swiftwasm/JavaScriptKit/pull/740 aimed to solve a similar problem, but was more narrow (and required more complex changes). While this approach requires more boilerplate for something like the `Polygon` case, it is also more general.
- In https://github.com/swiftwasm/JavaScriptKit/issues/737, @krodak suggested using a custom code-generation tool. While this is a possible approach, implementing this correctly is tricky and it makes sense to me for BridgeJS to have built-in support for what is likely going to be a common problem.
- Not doing anything: the status quo of requiring users to modify their Swift code to be less idiomatic or create a duplicate hierarchy of types. This would likely be an impediment to adopting BridgeJS, especially in larger projects who may have more consumers than just JavaScript.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
@@ -958,7 +958,7 @@ extension BridgeType {
         case .array, .dictionary:
             return LoweringParameterInfo(loweredParameters: [])
         case .alias:
-            preconditionFailure()
+            preconditionFailure("`.alias` must be resolved by `.unaliased` before reaching loweringParameterInfo")
         }
     }
 
@@ -1032,7 +1032,7 @@ extension BridgeType {
         case .array, .dictionary:
             return LiftingReturnInfo(valueToLift: nil)
         case .alias:
-            preconditionFailure()
+            preconditionFailure("`.alias` must be resolved by `.unaliased` before reaching liftingReturnInfo")
         }
     }
 }
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
@@ -172,8 +172,7 @@ public struct ImportTS {
             if loweringInfo.useBorrowing {
                 let returnVariableName = "ret\(borrowedArguments.count)"
                 let assign = needsReturnVariable ? "let \(returnVariableName) = " : ""
-                let loweredAlias = param.type.lowerAliases(expression: param.name)
-                body.write("\(assign)\(loweredAlias).bridgeJSWithLoweredParameter { \(pattern) in")
+                body.write("\(assign)\(param.name).bridgeJSWithLoweredParameter { \(pattern) in")
                 body.indent()
                 borrowedArguments.append(
                     BorrowedArgument(
@@ -204,8 +203,7 @@ public struct ImportTS {
                         "(\(raw: param.name) as! _BridgedSwiftProtocolExportable).bridgeJSLowerAsProtocolReturn()"
                     )
                 } else {
-                    let loweredAlias = param.type.lowerAliases(expression: param.name)
-                    initializerExpr = ExprSyntax("\(raw: loweredAlias).bridgeJSLowerParameter()")
+                    initializerExpr = ExprSyntax("\(raw: param.name).bridgeJSLowerParameter()")
                 }
 
                 if loweringInfo.loweredParameters.isEmpty {
@@ -296,21 +294,18 @@ public struct ImportTS {
 
             if returnType.usesSideChannelForOptionalReturn() {
                 // Side channel returns: extern function returns Void, value is retrieved via side channel
-                let liftCall = "\(returnType.unaliased.swiftType).bridgeJSLiftReturnFromSideChannel()"
-                body.write("return \(returnType.liftAliases(expression: liftCall))")
+                body.write("return \(returnType.swiftType).bridgeJSLiftReturnFromSideChannel()")
             } else {
                 let liftExpr: String
                 switch returnType {
                 case .closure(let signature, _):
                     liftExpr = "_BJS_Closure_\(signature.mangleName).bridgeJSLift(ret)"
                 default:
-                    let liftCall: String
                     if liftingInfo.valueToLift != nil {
-                        liftCall = "\(returnType.unaliased.swiftType).bridgeJSLiftReturn(ret)"
+                        liftExpr = "\(returnType.swiftType).bridgeJSLiftReturn(ret)"
                     } else {
-                        liftCall = "\(returnType.unaliased.swiftType).bridgeJSLiftReturn()"
+                        liftExpr = "\(returnType.swiftType).bridgeJSLiftReturn()"
                     }
-                    liftExpr = returnType.liftAliases(expression: liftCall)
                 }
                 body.write("return \(liftExpr)")
             }
@@ -908,7 +903,7 @@ extension BridgeType {
     }
 
     func loweringParameterInfo(context: BridgeContext = .importTS) throws -> LoweringParameterInfo {
-        switch self {
+        switch self.unaliased {
         case .bool: return .bool
         case .integer(let t): return LoweringParameterInfo(loweredParameters: [("value", t.wasmCoreType)])
         case .float: return .float
@@ -962,8 +957,8 @@ extension BridgeType {
             return LoweringParameterInfo(loweredParameters: params, useBorrowing: wrappedInfo.useBorrowing)
         case .array, .dictionary:
             return LoweringParameterInfo(loweredParameters: [])
-        case .alias(_, let underlying):
-            return try underlying.loweringParameterInfo(context: context)
+        case .alias:
+            preconditionFailure()
         }
     }
 
@@ -983,7 +978,7 @@ extension BridgeType {
     func liftingReturnInfo(
         context: BridgeContext = .importTS
     ) throws -> LiftingReturnInfo {
-        switch self {
+        switch self.unaliased {
         case .bool: return .bool
         case .integer(let t): return LiftingReturnInfo(valueToLift: t.wasmCoreType)
         case .float: return .float
@@ -1026,7 +1021,7 @@ extension BridgeType {
         case .nullable(let wrappedType, _):
             // jsObject and `@JS struct` use the stack ABI for optionals — the thunk returns
             // void and the value (plus isSome discriminator) flows through the stacks.
-            if case .jsObject = wrappedType.unaliased {
+            if case .jsObject = wrappedType {
                 return LiftingReturnInfo(valueToLift: nil)
             }
             if case .swiftStruct = wrappedType, context == .importTS {
@@ -1036,8 +1031,8 @@ extension BridgeType {
             return LiftingReturnInfo(valueToLift: wrappedInfo.valueToLift)
         case .array, .dictionary:
             return LiftingReturnInfo(valueToLift: nil)
-        case .alias(_, let underlying):
-            return try underlying.liftingReturnInfo(context: context)
+        case .alias:
+            preconditionFailure()
         }
     }
 }
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
@@ -172,7 +172,8 @@ public struct ImportTS {
             if loweringInfo.useBorrowing {
                 let returnVariableName = "ret\(borrowedArguments.count)"
                 let assign = needsReturnVariable ? "let \(returnVariableName) = " : ""
-                body.write("\(assign)\(param.name).bridgeJSWithLoweredParameter { \(pattern) in")
+                let loweredAlias = param.type.lowerAliases(expression: param.name)
+                body.write("\(assign)\(loweredAlias).bridgeJSWithLoweredParameter { \(pattern) in")
                 body.indent()
                 borrowedArguments.append(
                     BorrowedArgument(
@@ -203,7 +204,8 @@ public struct ImportTS {
                         "(\(raw: param.name) as! _BridgedSwiftProtocolExportable).bridgeJSLowerAsProtocolReturn()"
                     )
                 } else {
-                    initializerExpr = ExprSyntax("\(raw: param.name).bridgeJSLowerParameter()")
+                    let loweredAlias = param.type.lowerAliases(expression: param.name)
+                    initializerExpr = ExprSyntax("\(raw: loweredAlias).bridgeJSLowerParameter()")
                 }
 
                 if loweringInfo.loweredParameters.isEmpty {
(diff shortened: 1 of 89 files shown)


Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSCodegenTests/Alias.swift
```diff
@@ -0,0 +1,389 @@
...
+            return Int32(1)
+        }
+    }
+}
+
+@_expose(wasm, "bjs_roundtripPolygon")
+@_cdecl("bjs_roundtripPolygon")
+public func _bjs_roundtripPolygon(_ polygon: UnsafeMutableRawPointer) -> UnsafeMutableRawPointer {
+    #if arch(wasm32)
+    let ret = roundtripPolygon(_: Polygon.bridgeFromJS(PolygonReference.bridgeJSLiftParameter(polygon)))
```
kateinoigakukun: Instead of calling `bridgeFromJS` / `bridgeToJS` in a call sequence directly, I think we can define `Polygon.bridgeJSLiftParameter`, then we don't need to add special handling in ExportedThunkBuilder and CallJSEmission, and also we can remove special conversions for container types (Dictionary, Array, Optional).

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -273,6 +273,7 @@ public enum BridgeType: Codable, Equatable, Hashable, Sendable {
     case swiftProtocol(String)
     case swiftStruct(String)
     indirect case closure(ClosureSignature, useJSTypedClosure: Bool)
+    indirect case alias(name: String, underlying: BridgeType)
```
kateinoigakukun, line 277: Do we really need to add `alias` case in BridgeType? If we generate conventional methods like `bridgeJSLiftParameter` and `bridgeJSLowerParameter`, we might be able to make BridgeType consumers agnostic to aliasness.

kateinoigakukun (commented):
Thank you for trying to tackle this and showing your idea with actual code! This is very interesting to me, and I just realized the ability to have different transferring representations solves some other problems like https://github.com/swiftwasm/JavaScriptKit/issues/496

Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSCodegenTests/Alias.swift
wfltaylor: Yeah, this makes a lot of sense

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
wfltaylor: How would you see aliased types represented? Using the `Polygon` example in the description and `@JS func roundtrip(_ polygon: Polygon) -> Polygon`, how would the `Polygon` be represented in the `ExportedFunction`? `ExportSwift` needs to know the Swift type and the type it bridges to, but `JSGlueGen` only needs to be aware of the bridged representation.

kateinoigakukun: Ah, good point. Then it's fine to keep the case.

Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSCodegenTests/Alias.swift
wfltaylor: I’ve pushed a commit which does this. My aim was for aliases to use the same ABI as the underlying type, which meant adding a fair number of conformances to `BridgeJSIntrinsics.swift`. This could be made simpler by relaxing this requirement.

Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
```diff
@@ -957,6 +957,8 @@ extension BridgeType {
             return LoweringParameterInfo(loweredParameters: params, useBorrowing: wrappedInfo.useBorrowing)
         case .array, .dictionary:
             return LoweringParameterInfo(loweredParameters: [])
+        case .alias:
+            preconditionFailure()
```
krodak: Both `.alias` cases in this extension (here in `loweringParameterInfo`, and the matching one in `liftingReturnInfo` just below) are unreachable today since both switch on `self.unaliased`. But a bare `preconditionFailure()` leaves nothing to debug if a future change to `unaliased` ever stops stripping `.alias`. A message turns a silent trap into a stated invariant:

```swift
case .alias:
    preconditionFailure("`.alias` must be resolved by `.unaliased` before reaching here")
```

Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
```diff
@@ -2680,6 +2685,8 @@ private extension BridgeType {
             return .stackABI
         case .nullable(let wrapped, _):
             return wrapped.optionalConvention
+        case .alias:
+            preconditionFailure()
```
krodak: Unlike the ImportTS lowering paths, this switches on `self`, not `.unaliased` - so an alias-wrapped optional (e.g. `Polygon?`) could reach this case and hit the bare `preconditionFailure()`. Delegating to the underlying type matches how `abiReturnType` / `mangleTypeName` / `requiresSideChannelForOptionalReturn` already handle `.alias`, and removes the latent crash even if every current caller happens to unalias first:

```swift
case .alias(_, let underlying):
    return underlying.optionalConvention
```

Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
```diff
@@ -2776,6 +2783,8 @@ private extension BridgeType {
             return []
         case .nullable(let wrapped, _):
             return wrapped.wasmParams
+        case .alias:
+            preconditionFailure()
```
krodak: Same as the `optionalConvention` case above - switches on `self`, so safer to delegate than to trap:

```swift
case .alias(_, let underlying):
    return underlying.wasmParams
```

Sources/JavaScriptKit/Macros.swift
```diff
@@ -113,8 +113,12 @@ public enum JSImportFrom: String {
 ///
 /// - Important: This feature is still experimental. No API stability is guaranteed, and the API may change in future releases.
 @attached(peer)
-public macro JS(namespace: String? = nil, enumStyle: JSEnumStyle = .const, identityMode: Bool = false) =
-    Builtin.ExternalMacro
+public macro JS(
+    as aliasOf: Any.Type? = nil,
+    namespace: String? = nil,
```
krodak, line 118: `@JS(as:)` is a meaningful new escape hatch and `_BridgedSwiftAlias` is public API, but there's no DocC article for it. Every other exporting capability has one under `Documentation.docc/Articles/BridgeJS/Exporting-Swift/` (Class, Static-Functions, Static-Properties...), and that's where people will look first. Worth one even while the feature is experimental - a stub to flesh out:

```
Documentation.docc/Articles/BridgeJS/Exporting-Swift/Exporting-Swift-Custom-Representation.md
- When to reach for @JS(as:) (copy-semantics cost, unsupported features)
- The bridgeToJS() / bridgeFromJS contract
- The Polygon / JSPolygon example from the PR description
- Supported underlying representations + current limits (no raw-value enums, protocols, chained aliases)
```

Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
```diff
@@ -1612,11 +1687,47 @@ private final class ExportSwiftAPICollector: SyntaxAnyVisitor {
         return true
     }
 
+    private func recordAlias(
+        node: some SyntaxProtocol & NamedDeclSyntax,
```
krodak, line 1741: `extractAliasTarget` short-circuits before namespace handling, so `@JS(as: X.self, namespace: "Foo")` silently drops the namespace. Since the macro accepts both arguments, a user could reasonably expect the alias to land in `Foo`. A diagnostic beats the silent drop - something like, at the top of `recordAlias`:

```swift
if parent.extractNamespace(from: jsAttribute) != nil {
    errors.append(DiagnosticError(
        node: node,
        message: "`namespace` is ignored on `@JS(as:)` types",
        hint: "Remove the `namespace:` argument; the alias adopts its target's representation"
    ))
}
```

(adjust to whatever namespace-extraction helper is in scope here)

Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
```diff
@@ -517,6 +534,57 @@ public final class SwiftToSkeleton {
...
+    fileprivate func extractAliasTarget(from jsAttribute: AttributeSyntax) -> TypeSyntax? {
+        guard
+            let arguments = jsAttribute.arguments?.as(LabeledExprListSyntax.self),
+            let asArg = arguments.first(where: { $0.label?.text == "as" }),
+            let memberAccess = asArg.expression.as(MemberAccessExprSyntax.self),
+            memberAccess.declName.baseName.text == "self",
+            let base = memberAccess.base
+        else {
+            return nil
+        }
```
krodak, line 546: If `as:` is present but not in `X.self` form, this returns nil and the declaration quietly falls through to normal `@JS` handling rather than reporting the malformed `as:`. Low odds given the `Any.Type?` signature, but it'd save someone a confusing "why isn't my alias working" later. Worth splitting "no `as:`" from "`as:` present but unparseable":

```swift
guard let asArg = arguments.first(where: { $0.label?.text == "as" }) else {
    return nil  // genuinely not an alias
}
guard let memberAccess = asArg.expression.as(MemberAccessExprSyntax.self),
    memberAccess.declName.baseName.text == "self",
    let base = memberAccess.base
else {
    // `as:` present but malformed - report instead of silently ignoring
    return nil
}
return TypeSyntax(stringLiteral: base.trimmedDescription)
```

Surfacing a diagnostic from here means threading an `errors` channel in, so this is a "if you think it's worth it" rather than a must.

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -1709,6 +1740,8 @@ extension BridgeType {
         case .dictionary(let valueType):
             // Dictionary mangling: "SD" prefix followed by value type (key is always String)
             return "SD\(valueType.mangleTypeName)"
+        case .alias(let name, _):
+            return "Al\(name.count)\(name)"
```
krodak, line 1775: Aliases mangle on name only (`Al<count><name>`) while the other cases fold in structural info. Fine as long as `name` stays the namespace-qualified `swiftCallName` and unique across modules - worth a one-liner so nobody later "simplifies" it to the bare type name:

```swift
case .alias(let name, _):
    // `name` is the namespace-qualified swiftCallName (unique); the underlying
    // representation isn't folded in because aliases bridge via their JS type's ABI.
    return "Al\(name.count)\(name)"
```

krodak (approved):
LGTM, just minor requests

kateinoigakukun (approved):
LGTM as long as @krodak's feedbacks are resolved.

krodak:
@wfltaylor hope you don't mind - I pushed a small commit to your branch to fold in the review feedback, and rebased on `main` so the snapshots line up with the recent changes there.

What I added:
- messages on the unreachable `.alias` precondition cases in ImportTS
- delegating the `.alias` cases in `optionalConvention`/`wasmParams` to the underlying type instead of trapping
- a diagnostic for `@JS(as:)` combined with `namespace:` (it was silently dropped before)
- a DocC article for the feature

I wanted to get it green so we can merge and start using it on our side (the `@JS(as: String.self)` shape is exactly what we need to drop a UUID shim). Thank you so much for the contribution, this is a really nice addition.

Plugins/BridgeJS/Tests/BridgeJSToolTests/Inputs/MacroSwift/Alias.swift
```diff
@@ -0,0 +1,121 @@
+@JS(as: PolygonReference.self) struct Polygon {
+    var vertices: [Double]
+
+    consuming func bridgeToJS() -> PolygonReference {
+        return PolygonReference(underlying: self)
+    }
+
+    static func bridgeFromJS(_ value: consuming PolygonReference) -> Polygon {
+        return value.underlying
+    }
```
MaxDesiatov, line 10: @krodak I have a concern that we duplicated these APIs of existing [`ConstructibleFromJSValue`](https://swiftpackageindex.com/swiftwasm/javascriptkit/main/documentation/javascriptkit/constructiblefromjsvalue)
[`ConvertibleToJSValue`](https://swiftpackageindex.com/swiftwasm/javascriptkit/main/documentation/javascriptkit/convertibletojsvalue) protocols? Is there a specific reason these weren't reused?

krodak: Good question. They look similar but sit on different layers, and reusing them would actually pull the dynamic path back in.

`ConvertibleToJSValue`/`ConstructibleFromJSValue` are the dynamic runtime FFI: the target is always `JSValue` (`var jsValue: JSValue`), and reconstruction is fallible (`construct(from:) -> Self?`) since a given `JSValue` may not match. That's the boxed, dynamically-checked path BridgeJS exists to avoid.

`_BridgedSwiftAlias` instead maps to an arbitrary bridged representation through an `associatedtype JSRepresentation`, and the generated glue rides that representation's own static, typed ABI: an `Int32` handle for a class, a single wasm scalar for a primitive, the string ABI for `String`, struct/enum stacks, and so on (the `extension _BridgedSwiftAlias where JSRepresentation: ...` overloads). Both directions are total and `consuming`, not fallible.

`JSValue` is supported too, but only as one possible representation (`@JS(as: JSValue.self)`), not the universal transport. Coupling this to `ConvertibleToJSValue` would force every aliased type through `JSValue` and lose the static typing that's the whole point, so keeping it as a separate, representation-generic protocol makes sense.

wfltaylor:
> @wfltaylor hope you don't mind - I pushed a small commit to your branch to fold in the review feedback, and rebased on `main` so the snapshots line up with the recent changes there.
> 
> What I added:
> 
> * messages on the unreachable `.alias` precondition cases in ImportTS
> * delegating the `.alias` cases in `optionalConvention`/`wasmParams` to the underlying type instead of trapping
> * a diagnostic for `@JS(as:)` combined with `namespace:` (it was silently dropped before)
> * a DocC article for the feature
> 
> I wanted to get it green so we can merge and start using it on our side (the `@JS(as: String.self)` shape is exactly what we need to drop a UUID shim). Thank you so much for the contribution, this is a really nice addition.

Thanks @krodak !!

PR #754, opened 2026-06-03
BridgeJS: Support optional @JSClass as exported function parameters

### LINKED ISSUE #751: BridgeJS: Optional @JSClass parameter fails lift/lower for generated bindings
## Summary
BridgeJS code generation fails when a `@JS` exported function takes an optional `@JSClass` parameter (e.g. `JSAbortSignal?`).

## Repro
In a BridgeJS-exported Swift API:

```swift
@JSClass
public struct JSAbortSignal {
    @JSFunction public func addEventListener(type: String, listener: JSTypedClosure<() -> Void>) throws(JSException)
}

@JS
public final class JSCanvasSession {
    @JS
    public func recognizeText(pageId: String, languageCode: String, abortSignal: JSAbortSignal?) async throws(JSException) -> String {
        return "{}"
    }
}
```

Run BridgeJS AoT generation + build.

## Actual
Generated code fails to compile in `Generated/BridgeJS.swift` with optional lift/lower errors for `Optional<JSAbortSignal>`, e.g.:

- `no exact matches in call to static method 'bridgeJSLiftParameter'`
- candidates require `Wrapped == JSObject` or protocol conformances not satisfied by `JSAbortSignal`

## Expected
Optional `@JSClass` params should be supported, or codegen should emit a clear diagnostic that this shape is unsupported.

krodak:
## Overview

Fixes #751. A `@JS` exported function taking (or returning) an optional `@JSClass` such as `JSAbortSignal?` failed to compile with `no exact matches in call to static method 'bridgeJSLiftParameter'` for `Optional<JSAbortSignal>`.

Optional `jsObject` (incl. `@JSClass`) parameters now use the direct `(isSome, objId)` ABI in both directions, matching plain `Optional<JSObject>` and exported `@JS class` parameters; only return values use the stack ABI.

**1. Export parameters: use the direct ABI.** `ExportSwift` no longer special-cases `.nullable(.jsObject)`, so exported parameters lift via `Optional<T>.bridgeJSLiftParameter(isSome, value)`. Adds the missing `Optional<_JSBridgedClass>.bridgeJSLiftParameter(isSome, objId)` and `bridgeJSLowerReturn()` so optional `@JSClass` values round-trip as both parameters and return values.

**2. Import parameters: align to the direct ABI.** `loweringParameterInfo` already lowered optional `jsObject` import parameters as direct `(isSome, value)`, but the JS lift still popped the object id off the stack that Swift never pushed, throwing `ReferenceError: Attempted to use invalid reference undefined` for a non-nil value. The JS lift now reads the object id directly, matching the extern.

A new `optionalParameterUsesStackABI` keeps the stack ABI only for the return direction of optional `jsObject`.

## Tests

- Export direction: end-to-end runtime test passing an optional imported `@JSClass` directly as an exported function parameter and return value (present and `null` cases), plus snapshot coverage for optional `JSObject?` and `@JSClass?`.
- Import direction: end-to-end runtime test passing an optional `JSObject?` into an imported JS function (present and `null` cases), plus updated snapshot coverage for the optional `@JSClass?` import parameter.
- Regenerated in-package snapshots and the AoT-committed bindings. Snapshot suites, the BridgeJS TypeScript declaration check, and the full wasm test suite pass.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
@@ -762,7 +762,7 @@ struct IntrinsicJSFragment: Sendable {
         }
 
         let innerFragment =
-            if wrappedType.optionalConvention == .stackABI {
+            if wrappedType.optionalParameterUsesStackABI {
                 try stackLowerFragment(elementType: wrappedType)
             } else {
                 try lowerParameter(type: wrappedType)
@@ -779,7 +779,7 @@ struct IntrinsicJSFragment: Sendable {
         kind: JSOptionalKind,
         innerFragment: IntrinsicJSFragment
     ) throws -> IntrinsicJSFragment {
-        let isStackConvention = wrappedType.optionalConvention == .stackABI
+        let isStackConvention = wrappedType.optionalParameterUsesStackABI
 
         return IntrinsicJSFragment(
             parameters: ["value"],
@@ -2693,6 +2693,24 @@ private extension BridgeType {
             return .stackABI
         case .nullable(let wrapped, _):
             return wrapped.optionalConvention
+        }
+    }
+
+    /// Whether an optional of this type pushes its payload onto the bridge stack
+    /// when passed as a *parameter*.
+    ///
+    /// This usually matches `optionalConvention == .stackABI`, but `jsObject`
+    /// optionals are the exception: their return values travel through the stack
+    /// while their parameters use the direct `(isSome, objId)` ABI, matching plain
+    /// `Optional<JSObject>` and exported `@JS class` parameters.
+    var optionalParameterUsesStackABI: Bool {
+        switch self {
+        case .jsObject:
+            return false
+        case .nullable(let wrapped, _):
+            return wrapped.optionalParameterUsesStackABI
+        default:
+            return optionalConvention == .stackABI
         }
     }
 
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
@@ -669,7 +669,7 @@ struct IntrinsicJSFragment: Sendable {
         }
 
         let innerFragment =
-            if wrappedType.optionalConvention == .stackABI {
+            if wrappedType.optionalParameterUsesStackABI {
                 try stackLiftFragment(elementType: wrappedType)
             } else {
                 try liftParameter(type: wrappedType, context: bridgeContext)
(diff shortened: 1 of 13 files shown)


kateinoigakukun:
Sorry, I'm still reviewing this, so please give me some more time :bow:

Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
```diff
@@ -1531,6 +1536,12 @@ extension BridgeType {
         case .unsafePointer: return .unsafePointer
         case .swiftProtocol: return .jsObject
         case .void: return .void
+        case .nullable(.jsObject, _):
+            // Optional jsObject (incl. @JSClass) uses the stack ABI: the presence
+            // flag and object id are transferred through the bridge stack rather
+            // than as direct ABI parameters. The lift expression is the no-argument
+            // `Optional<T>.bridgeJSLiftParameter()` that pops them off the stack.
+            return LiftingIntrinsicInfo(parameters: [])
```
kateinoigakukun: I think we should use the direct `isSome, objId` on parameter ABI as well as `Optional<JSObject>` instead of stack ABI here.

kateinoigakukun (changes requested):

kateinoigakukun (approved):

Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
krodak: Thanks, that makes sense, the import/export optional conventions still trip me up a bit 😅 
I've pushed an update that keeps optional `jsObject/@JSClass` parameters on the direct `(isSome, objId)` ABI (adding the missing `Optional<_JSBridgedClass>.bridgeJSLiftParameter(isSome, objId)` overload), with only the return value going through the stack.

While doing this I've noticed what looks like a related discrepancy on the import side - for a scalar optional `JSObject?/@JSClass?` passed as a parameter to an imported JS function, the two sides seem to disagree:

- `loweringParameterInfo` is direct - .jsObject lowers to `[isSome, value]`

- JS lift `optionalLiftParameter` still uses the stack convention (optionalConvention == .stackABI), so it reads `isSome directly but i32Stack.pop()s the object id that Swift never pushed.

The null case works (nothing is popped), but passing a non-nil value throws , it just wasn't caught before because there's no runtime test for a scalar optional `JSObject?/@JSClass?` import parameter (the existing optional-import coverage is I`nt?/String?/JSValue-array/dictionary`).

Aligning the import side to the direct (isSome, objId) ABI too (matching the extern and the export side) fixes it. I've pushed this to second commit and updated PR description, not worth yet another PR I guess 😄

krodak: Sorry, was pretty late with posting that comment, but if CI is green and there is approval, I'll assume it's fine 😉

PR #755, opened 2026-06-08
BridgeJS: Support optional @JS struct in imported function signatures

krodak:
## Overview

Optional `@JS struct`s could not be used as parameters or return values of imported (`@JSFunction`) signatures. While an optional primitive works, an optional `@JS struct` did not:

```swift
@JS struct Point { var x: Int; var y: Int }

@JSFunction func roundTrip(_ point: Point?) throws(JSException) -> Point? // did not compile
```

The code generator lowered `Optional<Point>` using the non-optional object-id ABI (`[isSome, objectId]` for parameters, a single `Int32` word for returns), for which no `Optional` lowering exists in the runtime, so the generated thunk failed to compile.

This change bridges optional `@JS struct`s through the stack ABI instead, an `isSome` discriminator plus the struct fields, exactly like optional arrays and dictionaries. `@JS struct`s already conform to the stack-based bridging protocols, so the runtime's `_BridgedAsOptional`/stack extensions and the JS link's stack handling already support this shape. Only the import-side lowering and lifting in the code generator needed to change:

**1. Parameter lowering.** `.nullable(.swiftStruct)` in the import context now lowers to a single `isSome` flag, with the struct transferred on the stack, rather than appending the non-optional `objectId` word.

**2. Return lifting.** The optional struct return is now lifted from the stack rather than from a single object-id word.

Generated thunk, before:

```swift
let (pointIsSome, pointObjectId) = point.bridgeJSLowerParameter() // no such overload
let ret = bjs_roundTrip(pointIsSome, pointObjectId)
return Optional<Point>.bridgeJSLiftReturn(ret)
```

after:

```swift
let pointIsSome = point.bridgeJSLowerParameter()
bjs_roundTrip(pointIsSome)
return Optional<Point>.bridgeJSLiftReturn()
```

The export side already used the stack ABI for optional structs and is unchanged. The change is guarded to the import context, so non-optional struct imports keep their existing object-id ABI.

Adds an end-to-end runtime round-trip test (`some` and `none`) and a codegen snapshot covering the new import shape.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ImportTS.swift
@@ -957,6 +957,10 @@ extension BridgeType {
             }
         case .namespaceEnum:
             throw BridgeJSCoreError("Namespace enums cannot be used as parameters")
+        case .nullable(.swiftStruct, _) where context == .importTS:
+            // Optional `@JS struct`s bridge through the stack (isSome discriminator + fields),
+            // like optional arrays/dictionaries, rather than the non-optional object-id ABI.
+            return LoweringParameterInfo(loweredParameters: [("isSome", .i32)])
         case .nullable(let wrappedType, _):
             let wrappedInfo = try wrappedType.loweringParameterInfo(context: context)
             var params = [("isSome", WasmCoreType.i32)]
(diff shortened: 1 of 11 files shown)


MaxDesiatov (approved):
Thanks!

Issue #753, opened 2026-06-02
BridgeJS: Support `@JS struct` on return type of async exported `@JS func`


kateinoigakukun:
https://swiftwasm.org/JavaScriptKit/PlayBridgeJS/?code=H4sIAAAAAAAAE6tWKi7PTCtRslJy8ApWKC4pKk0uUXDLz1eoro3JAwmlleYlK6Snlrjl52toKiQWV%2BYlK%2BjaQZUo6SillBQrWSkp1QIAGbFRnEkAAAA%3D&enc=gzip-b64

Input:

```swift
@JS struct Foo {}
@JS func getFoo() async -> Foo {}
```

Expected Output:

```swift
@JSFunction func Promise_resolve_<mangled Foo>(_ promise: JSObject, _ value: Foo) throws(JSException)
@JSFunction func Promise_reject(_ promise: JSObject, _ value: JSValue) throws(JSException)

@_expose(wasm, "bjs_getFoo")
@_cdecl("bjs_getFoo")
public func _bjs_getFoo() -> Int32 {
    #if arch(wasm32)
    return _bjs_makePromise(resolve: Promise_resolve_<mangled Foo> reject: Promise_reject) {
        return await getFoo()
    }
    #else
    fatalError("Only available on WebAssembly")
    #endif
}
```

Actual Output:

```swift
@_expose(wasm, "bjs_getFoo")
@_cdecl("bjs_getFoo")
public func _bjs_getFoo() -> Int32 {
    #if arch(wasm32)
    let ret = JSPromise.async {
        return await getFoo().jsValue
    }.jsObject
    return ret.bridgeJSLowerReturn()
    #else
    fatalError("Only available on WebAssembly")
    #endif
}
```

krodak:
https://github.com/swiftwasm/JavaScriptKit/pull/758 closed

PR #758, opened 2026-06-08
BridgeJS: Support non-ConvertibleToJSValue async exported return types

krodak:
## Overview

Addresses #753. Async exported functions previously required a return type conforming to `ConvertibleToJSValue`. The generated thunk wrapped the body in `JSPromise.async { ... }` and lowered the result with `.jsValue`, so types that have no `.jsValue` representation (`@JS struct`s, raw-value and case enums, and their `Optional`/`Array`/`Dictionary` compositions) could not be returned from an `async @JS func`:

```swift
@JS struct Point { var x: Int; var y: Int }
@JS func getPoint() async -> Point { ... }   // did not compile: Point has no `.jsValue`
```

This PR settles every `async` exported return through a single new mechanism that lowers the value via the regular imported-parameter ABI, so all bridged types and their nested compositions can be returned.

## Approach

**1. Create the Promise eagerly, settle it from a `Task`.**
An exported async function must hand a value back to JavaScript synchronously (the Wasm export returns an `i32` object id), while the work itself is asynchronous. The thunk therefore creates a JS `Promise` immediately via a new `_bjs_makePromise` intrinsic, returns its retained object id, and resolves or rejects it later from a `Task`:

```swift
// generated thunk (struct return)
public func _bjs_getPoint() -> Int32 {
    return _bjs_makePromise(resolve: Promise_resolve_5PointV, reject: Promise_reject) {
        return await getPoint()
    }
}
```

The `resolve`/`reject` settlers are stashed on the `Promise` under a `Symbol` (rather than plain string fields) so they cannot clash with anything else on the object.

**2. Settle through generated `@JSFunction` thunks rather than ad-hoc lowering.**
`_bjs_makePromise` lives in the JavaScriptKit library and cannot name the per-module, per-type lowering that the generator produces. So the generator emits a `Promise_resolve_<mangled>` thunk per distinct return type (deduplicated by mangled name) and a single shared `Promise_reject`, and injects them into `_bjs_makePromise`. Each resolve thunk lowers its value through the **standard imported-parameter ABI**:

```swift
@JSFunction func Promise_resolve_5PointV(_ promise: JSObject, _ value: Point) throws(JSException)
@JSFunction func Promise_reject(_ promise: JSObject, _ value: JSValue) throws(JSException)
```

The benefit of reusing the imported-parameter ABI is that **every existing per-type lowering is inherited for free**: struct fields pushed on the stack, enum tags, optional `isSome` discriminators, arrays, and dictionaries all round-trip with no new marshaling code. `reject` uses the full `JSValue` ABI so any thrown error is forwarded faithfully.

**3. One path for all async returns.**
Rather than branch between `JSPromise.async` + `.jsValue` for `ConvertibleToJSValue` returns and `_bjs_makePromise` for the rest, every async return now settles through `_bjs_makePromise`. This removes the dual codegen path and the type predicate that used to select between them. `Void` is the one special case: a `Void` value can't cross the bridge as a parameter, so its `resolve` thunk takes only the promise (via a small `_bjs_makePromise` overload) and settles with `undefined`.

**4. Drain stack parameters synchronously before the deferred `Task`.**
Complex parameters travel via a single shared, mutable bridge stack. Because the async body runs later on a `Task`, any stack-using parameter must be lifted into a local **in the thunk** before the `Task` is scheduled, otherwise an interleaved bridge call would corrupt the shared stack. The thunk hoists these lifts in reverse (LIFO) order to preserve the stack discipline:

```swift
public func _bjs_combine() -> Int32 {
    let _tmp_b = AsyncPoint.bridgeJSLiftParameter()
    let _tmp_a = AsyncPoint.bridgeJSLiftParameter()
    return _bjs_makePromise(resolve: ..., reject: ...) {
        return await combine(_: _tmp_a, _: _tmp_b)
    }
}
```

**5. Annotate the throwing async closure explicitly.**
A throwing async body needs an explicit `() async throws(JSException) -> T in` closure type; without it Swift infers `throws(any Error)` instead of `throws(JSException)`. Non-throwing bodies infer correctly and are left unannotated.

**6. Diagnose, rather than miscompile, the unsupported cases.**
The few types that cannot be lowered through the imported-parameter ABI (associated-value enums, protocols, namespace enums, including those nested in `Optional`/`Array`/`Dictionary`) are rejected with a clear BridgeJS diagnostic instead of producing uncompilable Swift.

## Tests

End-to-end runtime round-trips plus codegen snapshots cover the matrix: `@JS struct`, case enum, and raw-value enum, each as a bare value, `Optional`, `Array`, and `Dictionary`; an `Int64`-backed enum (base and optional, exercising the `i64` resolve-helper signature); `Void` and `ConvertibleToJSValue` returns; the throwing variant, the reject path (asserted via `assert.rejects`), multi-parameter stack hoisting, a rich nested struct, an async class method, and concurrent calls via `Promise.all` (the regression guard for shared-stack corruption). Unsupported types are covered by the diagnostic path.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -90,9 +90,75 @@ public class ExportSwift {
                 decls.append(contentsOf: try renderSingleExportedClass(klass: klass))
             }
         }
+
+        try withSpan("Render Async Promise Helpers") { [self] in
+            let asyncResolveTypes = skeleton.asyncPromiseResolveReturnTypes
+            if !asyncResolveTypes.isEmpty {
+                decls.append(contentsOf: try renderPromiseRejectHelper())
+                for type in asyncResolveTypes {
+                    decls.append(contentsOf: try renderPromiseResolveHelper(type))
+                }
+            }
+        }
         return withSpan("Format Export Glue") {
             return decls.map { $0.description }.joined(separator: "\n\n")
         }
+    }
+
+    /// Generates the per-type `Promise_resolve_<mangled>` settlement helper.
+    private func renderPromiseResolveHelper(_ type: BridgeType) throws -> [DeclSyntax] {
+        try renderPromiseSettleHelper(
+            functionName: "Promise_resolve_\(type.mangleTypeName)",
+            externName: "promise_resolve_\(moduleName)_\(type.mangleTypeName)",
+            valueType: type
+        )
+    }
+
+    /// Generates the shared `Promise_reject` settlement helper.
+    private func renderPromiseRejectHelper() throws -> [DeclSyntax] {
+        try renderPromiseSettleHelper(
+            functionName: "Promise_reject",
+            externName: "promise_reject_\(moduleName)",
+            valueType: .jsValue
+        )
+    }
+
+    /// Generates a `@JSFunction func <functionName>(_ promise: JSObject, _ value: T)` and its
+    /// glue, lowering `value` through the standard imported-parameter ABI.
+    private func renderPromiseSettleHelper(
+        functionName: String,
+        externName: String,
+        valueType: BridgeType
+    ) throws -> [DeclSyntax] {
+        let effects = Effects(isAsync: false, isThrows: true)
+        // `Void` can't cross the bridge as a parameter, so the void helper takes only the promise.
+        var parameters = [Parameter(label: nil, name: "promise", type: .jsObject(nil))]
+        if valueType != .void {
+            parameters.append(Parameter(label: nil, name: "value", type: valueType))
+        }
+        let builder = try ImportTS.CallJSEmission(
+            moduleName: "bjs",
+            abiName: externName,
+            effects: effects,
+            returnType: .void,
+            context: .importTS
+        )
+        for parameter in parameters {
+            try builder.lowerParameter(param: parameter)
+        }
+        try builder.call()
+        try builder.liftReturnValue()
+
+        let valueParam = valueType == .void ? "" : ", _ value: \(valueType.swiftType)"
+        let macroDecl: DeclSyntax =
+            "@JSFunction func \(raw: functionName)(_ promise: JSObject\(raw: valueParam)) throws(JSException)"
+        let glueDecl = builder.renderThunkDecl(
+            name: "_$\(functionName)",
+            parameters: parameters,
+            returnType: .void,
+            effects: effects
+        )
+        return [macroDecl, builder.renderImportDecl(), glueDecl]
     }
 
     class ExportedThunkBuilder {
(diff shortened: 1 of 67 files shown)


krodak:
@kateinoigakukun let me know if approach aligns with your intention - if not just let me know which parts and I'll update accordingly 🙏🏻

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -1584,6 +1608,23 @@ extension BridgeType {
         return false
     }
 
+    /// Whether an `async` return of this type settles through `_bjs_makePromise` rather than
+    /// the `.jsValue` path (the non-`ConvertibleToJSValue` types the bridged-parameter ABI can transfer).
+    public var usesAsyncPromiseResolve: Bool {
```
kateinoigakukun: I think we don't need to keep `.jsValue` path. Removing it would simplify the codegen.

Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
```diff
@@ -526,6 +547,38 @@ public struct BridgeJSLink {
                     }
                 }
 
+                // Always provided: the runtime's `_bjs_makePromise` imports it unconditionally.
+                printer.write("bjs[\"swift_js_make_promise\"] = function() {")
+                printer.indent {
+                    printer.write("let resolve, reject;")
+                    printer.write("const promise = new Promise((res, rej) => { resolve = res; reject = rej; });")
+                    printer.write("promise.__bjs_resolve = resolve;")
```
kateinoigakukun: Maybe better to consider using Symbol field instead of just writing new fields just in case.

kateinoigakukun (commented):
Overall, looks good to me! Thanks!

Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
```diff
@@ -200,6 +272,9 @@ public class ExportSwift {
             }
 
             if effects.isAsync, returnType != .void {
+                if asyncResolveReturnType != nil {
+                    return CodeBlockItemSyntax(item: .init(StmtSyntax("return \(raw: callExpr)")))
+                }
                 return CodeBlockItemSyntax(item: .init(StmtSyntax("return \(raw: callExpr).jsValue")))
```
kateinoigakukun:
```suggestion
                return CodeBlockItemSyntax(item: .init(StmtSyntax("return \(raw: callExpr)")))
```
Now `.jsValue` path is deadcode, I believe.

Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
```diff
@@ -506,8 +596,23 @@ public class ExportSwift {
         return decls
     }
 
+    private func configureAsyncResolve(
```
kateinoigakukun: This is easy to forget to call, can we move this inside `ExportedThunkBuilder.init`?

kateinoigakukun (approved):

PR #760, opened 2026-06-10
BridgeJS: Fix reject path of zero-parameter async throwing exports

krodak:
## Overview

A zero-parameter `async throws(JSException)` export traps the Wasm instance when it throws. The generated thunk wraps the call in a `_bjs_makePromise` body closure; with no parameters there is nothing to capture, so the closure lowers via `thin_to_thick_function`, and invoking that value with an indirect typed error miscompiles on Wasm (swiftlang/swift#89320): the thrown `JSException` arrives corrupted in `Promise_reject` and lowering it traps with `RuntimeError: table index is out of bounds` (or the promise rejects with a garbage value). Exports with parameters are unaffected because their hoisted parameter locals are captured, which makes the body closure a partial apply with a matching call signature.

```swift
@JS func ping() async throws(JSException) -> String {
    throw JSException(JSError(message: "failure").jsValue)
}
```

Generated thunk before:

```swift
return _bjs_makePromise(resolve: Promise_resolve_SS, reject: Promise_reject) { () async throws(JSException) -> String in
    return try await ping()
}
```

After:

```swift
let __bjs_capture = 0
return _bjs_makePromise(resolve: Promise_resolve_SS, reject: Promise_reject) { [__bjs_capture] () async throws(JSException) -> String in
    _ = __bjs_capture
    return try await ping()
}
```

**1. Force a capture in captureless async throwing thunk bodies.** When the body closure would otherwise capture nothing (zero-parameter top-level functions, static methods, and constructors share this shape), the generator emits a dummy local captured by the closure. The body must also read the captured value: an unread capture list entry is dropped by capture analysis, leaving the closure thin and the bug in place. The read is what produces a real `partial_apply`.

**2. Tests.** A codegen snapshot covers the zero-parameter shape, and an end-to-end regression test asserts the rejection arrives with the thrown message instead of trapping. Existing thunks are emitted byte-identically.

This is a codegen-level workaround for swiftlang/swift#89320; once the IRGen fix in swiftlang/swift#89715 lands in a release toolchain, the forced capture can be removed.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -426,21 +426,45 @@ public class ExportSwift {
         /// A throwing async body needs an explicit closure type, otherwise Swift infers
         /// `throws(any Error)` instead of `throws(JSException)`.
         /// See: https://github.com/swiftlang/swift/issues/76165
-        private func asyncThrowsClosureHead(returnSpelling: String?) -> String {
+        private func asyncThrowsClosureHead(returnSpelling: String?, forcesCapture: Bool) -> String {
             guard effects.isThrows else { return "" }
             let returns = returnSpelling.map { " -> \($0)" } ?? ""
-            return " () async throws(JSException)\(returns) in"
+            let capture = forcesCapture ? "[__bjs_capture] " : ""
+            return " \(capture)() async throws(JSException)\(returns) in"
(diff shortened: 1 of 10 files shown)


kateinoigakukun (approved):
This is fine right now, can you create an issue on JSKit repo to track this FIXME item?

krodak:
@kateinoigakukun sure, done: https://github.com/swiftwasm/JavaScriptKit/issues/761

Issue #761, opened 2026-06-10
BridgeJS: Remove forced capture workaround in async thunk body closures once swiftlang/swift#89715 lands


krodak:
Tracks removal of the workaround introduced in #760.

## Context

A zero-parameter `async throws(JSException)` export generates a captureless `_bjs_makePromise` body closure, which lowers via `thin_to_thick_function` and miscompiles on Wasm (swiftlang/swift#89320): the thrown error is corrupted across the async unwind, trapping in `Promise_reject` or rejecting with a garbage value. #760 works around this at the codegen level by forcing a capture in the emitted body closure (the capture must also be read in the body, since an unread capture list entry is dropped by capture analysis and the closure stays thin):

```swift
let __bjs_capture = 0
return _bjs_makePromise(resolve: Promise_resolve_SS, reject: Promise_reject) { [__bjs_capture] () async throws(JSException) -> String in
    _ = __bjs_capture
    return try await ping()
}
```

## What to do once the compiler fix ships

The proper fix is swiftlang/swift#89715 (IRGen: fix async typed throws miscompiles on Wasm). Once it is available in the oldest Swift toolchain JavaScriptKit supports:

- Remove `asyncThrowsBodyForcesCapture` and the forced-capture emission from `ExportedThunkBuilder` in `Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift`.
- Regenerate snapshots and the AoT-committed bindings.
- Keep the `zeroArgAsyncThrows` end-to-end regression test as the guard that the reject path still works without the workaround.

Issue #537, opened 2026-01-28
[BridgeJS] Support throws/async for closures


kateinoigakukun:

krodak:
Shipped in #766

PR #766, opened 2026-06-11
BridgeJS: Support throws and async for closures

krodak:
## Overview

Adds `throws(JSException)` and `async` (and `async throws(JSException)`) support to BridgeJS-bridged closures, in both directions: closures Swift receives from JavaScript, and closures Swift hands to JavaScript. Previously closures were limited to synchronous, non-throwing signatures (#537).

Async settlement reuses the `_bjs_makePromise` + per-type `Promise_resolve_<mangled>` / `Promise_reject` machinery introduced in #758, so every bridged return type is inherited for free, including the associated-value enums added in #764; no parallel mechanism is introduced.

**1. Effect-aware closure mangling.** Closure ABI symbol names encode effects with the Swift ABI operators (`Ya` for async, `K` for throws), so signatures that differ only by effects no longer collide.

**2. Throwing closures (both directions).** A JavaScript callback's thrown `JSException` propagates into Swift; a Swift closure's thrown `JSException` is routed back to JavaScript. Only `throws(JSException)` is supported; plain `throws` is diagnosed.

**3. Async closures (both directions).** A JS-to-Swift callback is awaited via `_bjs_awaitPromise`; a Swift-to-JS closure returns a `Promise` settled via `_bjs_makePromise`. TypeScript surfaces these as `(args) => Promise<R>`. Supported return types mirror async functions, including `@JS struct`, enums (associated-value included), and `Optional`/`Array`/`Dictionary` compositions; the remaining unsupported async return types (protocols, namespace enums) are diagnosed rather than miscompiled.

**4. Lifetime.** The closure box survives across suspension without an explicit pin: the extracted closure value, captured by the settling `Task`, keeps it alive even if JavaScript releases the closure mid-flight. Covered by release-race and concurrent-invocation tests.

## Commit structure

The work is split so the parts can be reviewed (or dropped) independently:

**Commit 1** is the feature itself, self-contained.

**Commit 2** adds a build-time warning for async throwing closures handed to JavaScript (introducing a warning severity to the BridgeJS diagnostics). It was written when the swiftlang/swift#89320 miscompile made that signature lose thrown errors at runtime; with commit 3 in place the warning no longer fires for a real risk, so this commit can be dropped, or kept if the warning-severity infrastructure is useful on its own.

**Commit 3** works around swiftlang/swift#89320 by boxing `JSException`'s stored properties into a private final class, shrinking the struct to a single stored reference so it travels in the direct typed-error convention and the miscompiled indirect-error path is never taken. The public API is unchanged; the cost is one heap allocation per thrown exception, on the error path only. With it, the previously gated Swift-to-JS `async throws` reject assertions run unconditionally and pass, and the `JSClosure.async` reject path gains its previously missing test. Removal once swiftlang/swift#89715 ships is tracked in #767 and is a library-internal layout change, not a breaking API change. Original exploration: https://github.com/PassiveLogic/JavaScriptKit/pull/13. The codegen-level counterpart for zero-parameter async throwing exports landed in #760 (tracked for removal in #761).

## Tests

Codegen snapshots and end-to-end runtime round-trips (both directions, throwing and async, resolve and reject, `Void`, `@JS struct`, associated-value enum returns, the release-race, and concurrent calls) pass.


kateinoigakukun (approved):
Thanks!

PR #768, opened 2026-06-12
BridgeJS: Normalize wasm pointer offsets in JS

krodak:
## Overview

This PR fixes signed `i32` pointer handling at the Swift-to-JavaScript boundary. WebAssembly pointers can legally point above the 2 GiB line, but the JS API exposes `i32` values as signed numbers. When generated BridgeJS glue or runtime helpers used those signed values directly as typed-array or `DataView` offsets, pointers above `0x7fff_ffff` could become negative and fail with an out-of-bounds `RangeError`.

The change normalizes pointer offsets with `>>> 0` before using them to index WebAssembly linear memory. It also canonicalizes Swift heap object wrapper identity keys so signed and unsigned representations of the same pointer do not miss the identity cache.

**1. BridgeJS generated glue**

Generated `swift_js_init_memory`, `swift_js_init_memory_with_result`, and released-closure debug paths now unsigned-normalize pointer inputs before creating `Uint8Array` views. Swift heap object wrappers normalize pointers before storing or looking up identity-cache entries.

**2. Runtime helpers**

Runtime `DataView` and typed-array helper paths now unsigned-normalize pointer offsets before reading or writing memory. This covers JSValue array decoding, object-ref decoding, result payload writes, string loading/decoding, and typed-array creation/loading.

### DIFF
--- a/Runtime/test/pointer-normalization.test.ts
+++ b/Runtime/test/pointer-normalization.test.ts
@@ -1,0 +1,114 @@
+import { describe, expect, test } from "vitest";
+import { SwiftRuntime } from "../src/index.js";
+import { decodeArray, decodeObjectRefs, Kind, write } from "../src/js-value.js";
+
+// A Wasm pointer is an `i32`. Once the linear memory grows past 2 GiB, valid
+// addresses set the high bit and the guest passes them to JavaScript as *negative*
+// numbers, which must be normalized with `>>> 0` before being used as memory
+// offsets. These tests use a real >2 GiB `WebAssembly.Memory` and a real
+// `SwiftRuntime`, exercising the actual code paths at such an address; without the
+// normalization each one throws a `RangeError` on the negative offset.
+const PAGE_SIZE = 64 * 1024;
+const HIGH_OFFSET = 0x8000_0000; // 2 GiB, the first address with the high bit set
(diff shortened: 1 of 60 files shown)


Plugins/BridgeJS/Tests/BridgeJSToolTests/BridgeJSLinkTests.swift
```diff
@@ -176,4 +176,25 @@ import Testing
         let bridgeJSLink = BridgeJSLink(skeletons: [outputSkeleton], sharedMemory: false)
         try snapshot(bridgeJSLink: bridgeJSLink, name: "IdentityModeClass.ConfigPointer")
     }
+
+    @Test
+    func normalizesPointersBeforeUsingThemInJavaScript() throws {
```
kateinoigakukun: I feel this test seems a little bit easy to be broken, maybe just a runtime test might be enough without this?

Runtime/test/pointer-normalization.test.ts
kateinoigakukun, line 1: Is it too hard write usual runtime test instead of mocked unit tests?

kateinoigakukun (approved):

Runtime/test/pointer-normalization.test.ts
krodak: Sorry, figured this out eventually 🙏🏻

PR #771, opened 2026-06-23
BridgeJS: Include Swift doc comments in generated d.ts

### LINKED ISSUE #769: BridgeJS: Include Swift API doc comments in generated d.ts
Input

```swift
/// API doc comment for eat
/// - Parameters:
///   - food: The food for the sloth to eat.
///   - quantity: The quantity of the food for the sloth to eat.
@JS func eat(_ food: String, quantity: Int) -> Int {}
```

Expected output (`bridge-js.d.ts`)

```ts
export type Exports = {
    /**
     * API doc comment for eat
     * @param food The food for the sloth to eat.
     * @param quantity: The quantity of the food for the sloth to eat.
     */
    eat(food: string, quantity: number): number;
}
```

krodak:
## Overview

Resolves #769. Swift API doc comments on exported declarations are now propagated into the generated TypeScript declarations as JSDoc, so editors surface hover documentation for the bridged API.

The exporter captures the leading doc comment (`///` and `/** */`) for each exported declaration into the skeleton, and the linker renders it as a single JSDoc block. The mapping is the inverse of the JSDoc to Swift DocC conversion the TS2Swift importer already performs, so the two directions stay symmetric:

- The leading description becomes the JSDoc body.
- `- Parameters:` / `- Parameter name:` items become `@param name <desc>`.
- `- Returns:` becomes `@returns <desc>`.
- `- Throws:` becomes `@throws <desc>`.

**1. Coverage.** Doc comments are emitted for exported functions, classes (description, constructor, methods, properties), structs (description, fields), enums (description and const-style object-type static members), protocols (description, methods, properties), and namespaced declarations in both the `Exports` type hierarchy and the `declare global { namespace ... }` declarations.

**2. Merged with default-value annotations.** The default-value JSDoc that was previously the only thing emitted is now folded into the same block, so a parameter never produces two separate comment blocks. Declarations without doc comments produce byte-identical output to before.

**3. Comment terminators are escaped.** A `*/` inside a Swift doc comment is rewritten so it cannot prematurely close the generated JSDoc block.

### Example

```swift
/// API doc comment for eat
/// - Parameters:
///   - food: The food for the sloth to eat.
///   - quantity: The quantity of the food.
@JS func eat(_ food: String, quantity: Int) -> Int { ... }
```

```ts
/**
 * API doc comment for eat
 * @param food The food for the sloth to eat.
 * @param quantity The quantity of the food.
 */
eat(food: string, quantity: number): number;
```

### Testing

Adds a `DocComments` codegen/link fixture covering the mapping (functions, classes, structs, enums, protocols, namespaced functions, default-value merge, and `*/` escaping), and documents a namespaced declaration in the existing `Namespaces` fixture to cover the `declare global` path. Snapshot, codegen, and generated-d.ts validation all pass.

### DIFF
--- a/Examples/PlayBridgeJS/Sources/PlayBridgeJS/Generated/JavaScript/BridgeJS.json
+++ b/Examples/PlayBridgeJS/Sources/PlayBridgeJS/Generated/JavaScript/BridgeJS.json
@@ -16,6 +16,7 @@
         "methods" : [
           {
             "abiName" : "bjs_PlayBridgeJS_updateDetailed",
+            "documentation" : "Structured entry point used by the playground so JS doesn't need to parse diagnostics.",
             "effects" : {
               "isAsync" : false,
               "isStatic" : false,
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
@@ -1266,8 +1266,54 @@ private final class ExportSwiftAPICollector: SyntaxAnyVisitor {
             returnType: returnType,
             effects: effects,
             namespace: finalNamespace,
-            staticContext: staticContext
+            staticContext: staticContext,
+            documentation: extractDocumentation(from: node)
         )
+    }
+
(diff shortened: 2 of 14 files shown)


kateinoigakukun (approved):
Thanks!

PR #782, opened 2026-06-23
CI: Build examples in parallel on pull requests

krodak:
## Overview

`build-examples` builds every example sequentially in a single job and takes ~57 minutes. Each example is its own SwiftPM package that recompiles JavaScriptKit and swift-syntax from scratch, so the cost is dominated by repeating that swift-syntax compile back-to-back.

This turns the pull-request `build-examples` job into a matrix - one job per example, running concurrently - so the wall-clock time drops to that of the slowest single example. `main` keeps the single `build-examples-deploy` job (full release build) plus the GitHub Pages deploy, so published artifacts are identical. Examples are built in `release` to match the deploy path, and each matrix step runs `build.sh` from the example directory via `working-directory` (mirroring `Utilities/build-examples.sh`).

## Timing

| setup | examples wall-clock |
|---|---|
| today (single sequential job) | ~57 min |
| parallel matrix | ~10 min |

Each example still takes ~10 min on its own (dominated by compiling swift-syntax from scratch), but they now run in parallel. The `main` deploy job is unchanged.

## Branch protection: required checks need updating

Because `build-examples` is now a matrix, the single `build-examples` status check is replaced by one per example. An admin needs to update the branch protection required checks for `main`: drop `build-examples` and add the six matrix contexts:

- `build-examples (ActorOnWebWorker)`
- `build-examples (Basic)`
- `build-examples (Embedded)`
- `build-examples (Multithreading)`
- `build-examples (OffscrenCanvas)`
- `build-examples (PlayBridgeJS)`

Use the job + matrix name exactly as above. The `Run unit tests /` prefix and the `(pull_request)` suffix shown in the PR merge box are display-only and are not part of the required-context string. Until this is done, the old `build-examples` context sits in "Expected — Waiting for status to be reported" and blocks merge.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -155,7 +155,44 @@ jobs:
             exit 1
           }
 
+  # Pull requests: compile every example in parallel just to catch breakage.
+  # One job per example avoids rebuilding JavaScriptKit + swift-syntax 6x in series,
+  # which collapses the wall-clock time from ~1h to that of the slowest single example.
   build-examples:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        example:
+          - ActorOnWebWorker
+          - Basic
+          - Embedded
+          - Multithreading
+          - OffscrenCanvas
+          - PlayBridgeJS
+    steps:
+      - uses: actions/checkout@v7
+      - uses: ./.github/actions/install-swift
+        with:
+          download-url: https://download.swift.org/development/ubuntu2204/swift-DEVELOPMENT-SNAPSHOT-2026-05-27-a/swift-DEVELOPMENT-SNAPSHOT-2026-05-27-a-ubuntu22.04.tar.gz
+      - uses: swiftwasm/setup-swiftwasm@v2
+        id: setup-wasm32-unknown-wasip1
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -155,7 +155,52 @@ jobs:
...
+      - uses: swiftwasm/setup-swiftwasm@v2
+        id: setup-wasm32-unknown-wasip1
+        with: { target: wasm32-unknown-wasip1 }
+      - uses: swiftwasm/setup-swiftwasm@v2
+        id: setup-wasm32-unknown-wasip1-threads
+        with: { target: wasm32-unknown-wasip1-threads }
+      - uses: actions/cache@v4
+        with:
+          path: |
+            Examples/${{ matrix.example }}/.build
```
kateinoigakukun: `.build` is not portable across machines, and easily get tricky issue, so I'd like to keep building from fresh .build.

kateinoigakukun (approved):

.github/workflows/test.yml
krodak: Makes sense - dropped the caching commit, so example builds always start from a fresh `.build`. Parallelism alone still takes `build-examples` from ~57 min to ~10 min wall-clock, which was the bulk of the win.

PR #783, opened 2026-06-30
BridgeJS: Fix issues exporting nested types

wfltaylor:
There are currently a few issues with the JS/TS code generation, due to a lack of merging. For example, the following Swift code:
```swift
@JS struct Control {
    @JS enum Variant: String { case button, slider }

    @JS struct Layout {
        @JS enum Alignment: String { case leading, trailing }
        var padding: Int
    }

    var name: String

    @JS init(name: String) { self.name = name }
}
```
generates a `.d.ts`:
```ts
export type Exports = {
    Control: {
        init(name: string): Control;
    },
    Control: {
        Variant: VariantObject,
        Layout: {
            Alignment: AlignmentObject
        }
    }
}
```

This doesn’t type check correctly. Similar issues occur with the emitted `.js`, except instead of failing to type check, these will cause runtime issues (since keys are overwritten instead of merged).

This PR is a bit of a refactor of `BridgeJSLink` to address this. Instead of being a member of a `NamespaceNode`, each class and struct becomes its own `NamespaceNode`. This makes it much easier to merge everything together so it can be emitted as one unit.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/ExportSwift.swift
@@ -560,7 +560,7 @@ public class ExportSwift {
         func callName(for property: ExportedProperty) -> String {
             switch self {
             case .enumStatic(let enumDef):
-                return property.callName(prefix: enumDef.swiftCallName)
+                return "\(enumDef.swiftCallName).\(property.name)"
             case .classStatic(let klass):
                 // property.callName() would use staticContext (the ABI name) as prefix;
                 // use swiftCallName directly so the emitted expression is valid Swift.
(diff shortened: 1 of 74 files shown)


krodak (commented):
Thanks for tackling this, @wfltaylor, one thing to address, I'll get back to you on it soon, good otherwise 👌🏻

krodak (approved):
Reviewed, looks good, merging the nested types this way is clean. 👍

Heads up: I pushed 2 commits on top. They're not strictly part of your change, but your new fixtures surfaced a pre-existing bug, so I folded them in:

- Fix static property call expression for nested structs/enums (`Outer_Inner.prop` → `Outer.Inner.prop`; the `.classStatic` case was already fixed this way, just extended it to struct/enum)`.
- Runtime coverage for the above via `NestedTypeHost`.

PR #786, opened 2026-07-08
BridgeJS: Unify optional stack encoding to presence-flag form

krodak:
## Overview

Unifies the stack encoding of `Optional` values to a single presence-flag form. Previously, associated-value enums were the one type with a specialized optional stack encoding: the presence signal was merged into the case-ID slot, with `-1` meaning `nil`. Every other type uses a `0`/`1` presence flag followed by the value. After this change, associated-value enums use the same flag form as everything else when crossing in stack positions (array elements, struct fields, dictionary values, nested enum payloads).

```
before:  caseId | -1            (one i32, meaning depends on the wrapped type)
after:   flag, then caseId      (flag is 0/1 for every type)
```

**1. Swift runtime.** The `bridgeJSStackPop/PushAsOptional` override on `_BridgedSwiftAssociatedValueEnum` is removed; the protocol defaults now serve every type. The direct return slot keeps the merged `caseId`/`-1` convention (it is a separate ABI position, shared with other enum returns); the previous implementation delegated to the stack helper, so the merged form is now inlined there with a comment.

**2. JS glue.** The associated-value enum special cases in `optionalElementRaiseFragment` / `optionalElementLowerFragment` are removed; associated-value enums flow through the same general path as all other wrapped types, which composes the existing per-type element fragments behind a presence flag.

## Cost

Optional associated-value enums in stack positions carry one extra `i32` (the flag) per crossing. Direct parameter and return conventions are unchanged.

## Motivation

This removes the only optional stack encoding that cannot be decoded without knowing the wrapped type: a popped `0` meant `nil` for most types but "case 0, payload follows" for associated-value enums. A type-agnostic consumer of the stack encoding needs one uniform optional form; this prepares for generic function support (#398), where a runtime codec decodes optionals without static knowledge of `T`. A follow-up PR adds that support.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
@@ -2195,35 +2195,6 @@ struct IntrinsicJSFragment: Sendable {
         wrappedType: BridgeType,
         kind: JSOptionalKind
     ) throws -> IntrinsicJSFragment {
-        if case .associatedValueEnum(let fullName) = wrappedType {
-            let base = fullName.components(separatedBy: ".").last ?? fullName
-            let absenceLiteral = kind.absenceLiteral
-            return IntrinsicJSFragment(
-                parameters: [],
-                printCode: { arguments, context in
-                    let (scope, printer) = (context.scope, context.printer)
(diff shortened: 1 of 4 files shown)


kateinoigakukun (approved):
Thanks!

PR #789, opened 2026-07-17
BridgeJS: Fix syntax error in generated JS

wfltaylor:
When lifting a string return from an instance method on a struct, the following invalid JavaScript could be generated:
```js
const ret = instance.exports.bjs_Vector2D_describe();
const ret = tmpRetString;
```

This PR addresses this using the existing scoped variable identifier disambiguator.


kateinoigakukun (approved):
Thanks! Would you mind adding a runtime test as well?

Issue #508, opened 2026-01-22
Support bundling of JS snippets


sliemeobn:
as mentioned in #477 we currently have no good solution for creating a library that contains both Swift and JS code.

even if we can automatically "import" JS globals via BridgeJS, certain cases will require JS code, eg when:
+ you need JS features that are not exposed through the bridge (eg: create a class)
+ the back and forth through the bridging layer adds unnecessary runtime work for a "chunk of JS work"

here is a use case from ElementaryUI:
I want to create Swift API that let's you define custom elements ("web components").
it is currently incubated in the elementary.codes repo ([ts code](https://github.com/elementary-swift/elementary.codes/blob/48e12277aadfafa855ffccfb0f1c8437b9ab5c17/elementary-components.ts), [swift code](https://github.com/elementary-swift/elementary.codes/blob/48e12277aadfafa855ffccfb0f1c8437b9ab5c17/swift/Sources/Snippets/_Components.swift))

ideally we can just develop a Swift package with JS functions and Swift functions, add and attribute, and it will all "just work" in the final app that uses this library.

PR #792, opened 2026-07-22
add JS-snippet support using `from: .module("/path/to/js-module.js")` syntax

sliemeobn:
allow to ship JS code as the source for imports (in addition to ".global" and explicitly provided imports)

```swift
@JSFunction(from: .module("/Modules/JSImportModule.mjs"))
func moduleAdd(_ lhs: Int, _ rhs: Int) throws(JSException) -> Int

@JSFunction(jsName: "renamedFunction", from: .module("/Modules/JSImportModule.mjs"))
func moduleRenamed() throws(JSException) -> String

@JSGetter(jsName: "version", from: .module("/Modules/JSImportModule.mjs"))
var moduleVersion: String
```

```js
// Modules/JSImportModule.mjs
export function moduleAdd(lhs, rhs) {
    return lhs + rhs;
}

export function renamedFunction() {
    return "renamed";
}

export const version = "1.0";
```

I noticed that currently there is no diagnostic when `from:` is used in places that do not support it, I did not add these checks (eg: class members, setters) to this PR more focussed. (https://github.com/swiftwasm/JavaScriptKit/issues/793)


sliemeobn:
closes #508

Plugins/PackageToJS/Sources/PackageToJS.swift
```diff
@@ -597,21 +604,40 @@ struct PackagingPlanner {
...
-                        try link.addSkeletonFile(data: data)
+                    try system.writeFile(
+                        atPath: scope.resolve(path: bridgeDts).path,
+                        content: Data(output.outputDts.utf8)
+                    )
+                }
+            )
+            packageInputs.append(
+                make.addTask(inputFiles: skeletons + [selfPath], output: bridgeModules) { _, scope in
+                    let output = try linkBridgeJS(scope)
```
krodak: Splitting this into a second task means `linkBridgeJS` (decode all skeletons + full codegen) runs twice on every build — even for the majority of projects with zero modules, which also end up with an empty `bridge-js-modules/` directory in their output package.

I get why the split is attractive: a separate task output means `cleanEverything` removes the directory. Memoizing the link result keeps that benefit without the double work — skeletons can't change within one `build()` invocation and MiniMake runs tasks sequentially, so the cache is safe:

```swift
var cachedLinkOutput: BridgeJSLinkOutput?
let linkBridgeJS: (MiniMake.VariableScope) throws -> BridgeJSLinkOutput = { scope in
    if let cachedLinkOutput { return cachedLinkOutput }
    var link = BridgeJSLink(
        sharedMemory: Self.isSharedMemoryEnabled(triple: triple)
    )
    for skeletonPath in skeletons {
        let url = URL(fileURLWithPath: scope.resolve(path: skeletonPath).path)
        try link.addSkeletonFile(data: Data(contentsOf: url))
    }
    let output = try link.link()
    cachedLinkOutput = output
    return output
}
```

Plugins/BridgeJS/Sources/BridgeJSLink/ImportedJSModuleRegistry.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    func configure(skeletons: [BridgeJSSkeleton]) throws {
+        aliases.removeAll(keepingCapacity: true)
+        artifacts.removeAll(keepingCapacity: true)
+
+        var sources: [Key: String] = [:]
+        for skeleton in skeletons {
+            for module in skeleton.imported?.modules ?? [] {
+                guard module.path.hasPrefix("/") else {
+                    throw BridgeJSLinkError(
```
krodak: `relativePath` is spliced straight into `outputDir.appending(path:)` in `PackagingPlanner`. `makeJavaScriptModuleFileLookup` guarantees no traversal for locally generated skeletons, so this is safe today — but skeletons also arrive from dependency packages' committed `BridgeJS.json`, and a path like `/../../evil.js` in one of those would write outside the output directory (and emit an escaping `import` line). Cheap to close here since we're already validating:

```swift
guard module.path.hasPrefix("/"),
    !module.path.split(separator: "/").contains("..")
else {
    throw BridgeJSLinkError(
        message: "JavaScript module path must start with '/' and must not contain '..': \(module.path)"
    )
}
```

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -1126,8 +1126,35 @@ private struct AsyncClosureReturnTypeCollector: BridgeSkeletonVisitor {
...
 /// - `global`: Read from `globalThis`.
-public enum JSImportFrom: String, Codable {
+public enum JSImportFrom: Codable, Equatable, Sendable {
     case global
+    case module(String)
+
+    public init(from decoder: any Decoder) throws {
+        let value = try decoder.singleValueContainer().decode(String.self)
+        self = value == "global" ? .global : .module(value)
+    }
```
krodak, line 1147: Encoding `.module` into the same single-value slot as `.global` is a nice backward-compatible trick, and the legacy-decode test covers it well. One downside: any unknown string (a future case, or a typo'd/hand-edited skeleton) silently decodes as `.module("whatever")`, and the failure surfaces much later as `Missing embedded JavaScript module for Foo/whatever`, which points nowhere useful. Since module paths are already required to start with `/`, the invalid case can fail loudly right at the boundary:

```swift
public init(from decoder: any Decoder) throws {
    let container = try decoder.singleValueContainer()
    let value = try container.decode(String.self)
    if value == "global" {
        self = .global
    } else if value.hasPrefix("/") {
        self = .module(value)
    } else {
        throw DecodingError.dataCorruptedError(
            in: container,
            debugDescription: "Unknown JSImportFrom value '\(value)'. Expected \"global\" or a '/'-rooted module path."
        )
    }
}
```

Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Importing-JavaScript/Importing-JS-Function.md
```diff
@@ -17,6 +17,20 @@ import JavaScriptKit
...
+export function add(a, b) { return a + b; }
+```
+
+```swift
+@JSFunction(from: .module("/JavaScript/math.js"))
+func add(_ a: Double, _ b: Double) throws(JSException) -> Double
+```
+
+The leading `/` denotes the Swift target root, not the filesystem root. BridgeJS copies explicitly referenced modules into the generated PackageToJS package. Multiple declarations may reference the same file; it is embedded and imported only once. `jsName` selects a differently named export, otherwise BridgeJS uses the normalized Swift name.
+
```
krodak, line 43: This PR itself had to add `"Modules"` to the `exclude:` list in `Package.swift`, and anyone following these docs will hit the same thing — SwiftPM's "found N file(s) which are unhandled" warning the moment they drop a `.js` file under their target. Worth calling out once here (and probably in the overview article too), e.g. right after this paragraph:

````markdown
SwiftPM does not know what to do with `.js`/`.mjs` files inside a target, so exclude the directory holding them to avoid an "unhandled files" warning:

```swift
.target(
    name: "MyApp",
    exclude: ["JavaScript"],
    plugins: [.plugin(name: "BridgeJS", package: "JavaScriptKit")]
)
```
````

File: Plugins/BridgeJS/Sources/BridgeJSPluginUtilities/InputDiscovery.swift [hunk 5]
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+/// JavaScript files must be declared as build-command inputs so SwiftPM reruns
+/// BridgeJS when a referenced module changes.
+func discoverJavaScriptModuleFiles(in directory: URL) -> [URL] {
+    guard
```
krodak: Every `.js`/`.mjs` anywhere under the target becomes a declared build-command input *and* is appended to the tool's argv, referenced or not. For a target that vendors JS assets (or has a stray `node_modules`), that's a lot of spurious rebuild triggers and a long command line — and a *directory* named `foo.js` currently passes the filter, only to fail later at `String(contentsOf:)` with a confusing error. The sibling `recursivelyCollectSwiftFiles` in `BridgeJSTool.swift` checks `isRegularFile`; doing the same here plus a small skip-list would cover it:

```swift
private let skippedDirectoryNames: Set<String> = ["node_modules"]

func discoverJavaScriptModuleFiles(in directory: URL) -> [URL] {
    guard
        let enumerator = FileManager.default.enumerator(
            at: directory,
            includingPropertiesForKeys: [.isRegularFileKey],
            options: [.skipsHiddenFiles]
        )
    else { return [] }

    var files: [URL] = []
    for case let fileURL as URL in enumerator {
        if skippedDirectoryNames.contains(fileURL.lastPathComponent) {
            enumerator.skipDescendants()
            continue
        }
        guard isJavaScriptModuleFile(fileURL),
            (try? fileURL.resourceValues(forKeys: [.isRegularFileKey]))?.isRegularFile == true
        else { continue }
        files.append(fileURL)
    }
    return files.sorted { $0.path < $1.path }
}
```

(`.build` and other dot-directories are already covered by `.skipsHiddenFiles`.)

Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Importing-JavaScript-into-Swift.md
```diff
@@ -25,10 +25,11 @@ Add the BridgeJS plugin and enable the Extern feature as described in <doc:Setti
 
 Create Swift declarations that mirror the JavaScript API you want to call.
 
-You can bring JavaScript into Swift in two ways:
+You can bring JavaScript into Swift in three ways:
```
krodak, line 29: Tiny one: the intro at line 11 of this article still says imports work in "two ways", so the article now disagrees with itself:

```markdown
You can import JavaScript APIs into Swift in three ways:
```

krodak (commented):
Nice feature — the design fits the existing patterns well and test coverage is strong. Details in the inline comments.

Two tiny things not worth their own threads: the registry error messages render a double slash (`TestModule//Modules/x.mjs` — path already starts with `/`), and a runtime test for a throwing module export would pin down the `setException` path too.

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
kateinoigakukun: We don't need to worry about backward compatibility for skeleton 👌

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
```diff
@@ -1126,8 +1126,35 @@ private struct AsyncClosureReturnTypeCollector: BridgeSkeletonVisitor {
...
+    public var modulePath: String? {
+        guard case .module(let path) = self else { return nil }
+        return path
+    }
+}
+
+/// A target-local ECMAScript module embedded into a BridgeJS skeleton.
+public struct ImportedJSModule: Codable, Equatable, Sendable {
+    public let path: String
+    public let source: String
```
kateinoigakukun: This is a bit concerning. Given that we commit skeleton files when we do AoT codegen, it's not ideal if we need to update skeleton files whenever we change snippet js files. I think we can include only relative path to the js files.

kateinoigakukun (commented):
IIUC the current approach requires developers to run the build plugins and link steps to reflect their changes against snippet js files into the final output, right? This is another challenge I wanted to explore to have faster iterations.

Plugins/BridgeJS/Sources/BridgeJSSkeleton/BridgeJSSkeleton.swift
sliemeobn: I started with "in-line" JS code snippets, then added file support, and then removed the "inline JS" snippets ; )

I somehow thought copying the file content in the skeleton file is a "cleaner" method of delivery - but I will try a version where we only store the file names and resolve files at link time.

Plugins/BridgeJS/Sources/BridgeJSLink/ImportedJSModuleRegistry.swift
sliemeobn: makes sense, will add that

Plugins/BridgeJS/Sources/BridgeJSPluginUtilities/InputDiscovery.swift
sliemeobn: I am not sure how common it is to have entire js packages inside Swift target folder (I really hope "not" ; ) Also, I am not so sure about hard-coding "node_modules". I'd rather leave that for a future improvement if that ever becomes a problem...

Plugins/PackageToJS/Sources/PackageToJS.swift
sliemeobn: based on @kateinoigakukun feedback below I will change the behavior to only use the path in the skeleton and adjust the link implementation accordingly - this should also address this.

Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Importing-JavaScript-into-Swift.md
sliemeobn: actually, IIUC, this refers to "either via macros, or via `bridge-js.d.ts` file.
I'll change it to `You can define JavaScript bindings for Swift in two ways:` to fit the rest of the doc better.

Plugins/BridgeJS/Sources/BridgeJSPluginUtilities/InputDiscovery.swift
sliemeobn: ok, I think since we no longer "fingerprint" or copy stuff into the skeleton, we can get rid of this entirely....

sliemeobn:
@krodak @kateinoigakukun I reworked the whole thing to not copy anything to the skeletons and only use the "from" paths during the PackageToJS / link run. I think this is better ultimately.

the only thing that is a bit worse is that I always load and decode all skeletons up-front to extract all js-files as mini-make inputs. I though about adding some caching or smarter loading, but I wanted to keep this PR more straight forward.

maybe a separate PR "optimize PackageToJS" would be a good idea anyway eventually - I am sure there are milliseconds to be found here and there.

Sources/JavaScriptKit/Documentation.docc/Articles/BridgeJS/Ahead-of-Time-Code-Generation.md
```diff
@@ -155,6 +155,8 @@ git add Sources/MyApp/Generated
 git commit -m "Update generated BridgeJS code"
 ```
 
+Changing only the contents of a JavaScript module referenced with `from: .module` does not require regenerating the Swift glue or skeleton. PackageToJS tracks the referenced file as an input and copies its current contents into the generated package on the next build.
```
kateinoigakukun: This is an implementation detail users don't need to worry about, so we can drop it from docs?

kateinoigakukun (approved):
Looks good to me!

sliemeobn:
any idea what causes the diff? I ran the generate stuff locally and get no changes - but CI has this error
```
diff --git a/Tests/BridgeJSRuntimeTests/Generated/BridgeJS.swift b/Tests/BridgeJSRuntimeTests/Generated/BridgeJS.swift
index 181224c..39de49c 100644
--- a/Tests/BridgeJSRuntimeTests/Generated/BridgeJS.swift
+++ b/Tests/BridgeJSRuntimeTests/Generated/BridgeJS.swift
@@ -17043,8 +17043,8 @@ fileprivate func bjs_moduleAdd_extern(_ lhs: Int32, _ rhs: Int32) -> Int32 {
 }
 
 func _$moduleAdd(_ lhs: Int, _ rhs: Int) throws(JSException) -> Int {
-    let lhsValue = lhs.bridgeJSLowerParameter()
     let rhsValue = rhs.bridgeJSLowerParameter()
+    let lhsValue = lhs.bridgeJSLowerParameter()
     let ret = bjs_moduleAdd(lhsValue, rhsValue)
     if let error = _swift_js_take_exception() {
         throw error
@@ -17179,8 +17179,8 @@ func _$ModuleCounter_value_get(_ self: JSObject) throws(JSException) -> Int {
 }
 
 func _$ModuleCounter_value_set(_ self: JSObject, _ newValue: Int) throws(JSException) -> Void {
-    let selfValue = self.bridgeJSLowerParameter()
     let newValueValue = newValue.bridgeJSLowerParameter()
+    let selfValue = self.bridgeJSLowerParameter()
     bjs_ModuleCounter_value_set(selfValue, newValueValue)
     if let error = _swift_js_take_exception() {
         throw error
```

PR #799, opened 2026-08-10
BridgeJS: Support generic functions on imported JS APIs

krodak:
## Summary

Import half of generic function support (#398), split out of #787 per review. An imported `@JSFunction` can take a type parameter, so one declaration covers every bridged type:

```swift
@JSFunction func parse<T: BridgedSwiftGenericBridgeable>(_ json: String) throws(JSException) -> T

let user: User = try parse(jsonString)   // T inferred from the call site
```

`T` can be any primitive, `String`, `JSValue`, or a `@JS` struct/enum/final class, bare or as `T?`, `[T]`, `[String: T]`. Also works on `@JSClass` inits, methods and statics, and for return-only generics. Exported generics are rejected with a diagnostic and land separately on top of this ABI.

## How it works

- Values cross on the existing stack ABI. The only new thing on the wire is an `i32` type ID per type parameter, so type-agnostic JS glue can pick the right lift/lower.
- Type IDs are pointer-based, not name-based (per #787 review): each type gets a `BridgeJSTypeHandle`, its address is the ID. Same-named types in different modules can't collide, and no existentials means it works under Embedded.
- Each module exports `bjs_<Module>_register_type_handles`, which hands its IDs to JS; JS pairs them with its codec table by index. Runs eagerly via a new `afterInitialize` instantiator hook, or lazily on first generic call (worker threads).
- Conformances are emitted for every `@JS` type, not just in modules that declare generics, module B can pass module A's type to a generic function (per #787 review).
- JS containers get one shared codec each (`__bjs_arrayCodec` etc.); the non-generic array/dict/optional paths now reuse them instead of inlining a copy per thunk (per #787 review on the stack ABI clones).

Builds without generics don't get the JS generic runtime, just a no-op registration hook per module with `@JS` types.

## Test plan

- [x] Codegen + link snapshots: free functions, inits, methods, statics, mixed/multiple params, return-only, wrapped forms
- [x] Diagnostics: import constraints, export rejection
- [x] `ImportGenericAPITests`: runtime round-trip of every bridgeable type through real JS, including `JSON.parse`
- [x] `Examples/Embedded` builds a generic round-trip in CI
- [x] Cross-module registration + runtime gating link tests


Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSCodegenTests/Alias.swift
```diff
@@ -394,4 +406,36 @@ func _$Surface_label_get(_ self: JSObject) throws(JSException) -> String {
...
+        UInt16.bridgeJSTypeID,
+        Int32.bridgeJSTypeID,
+        UInt32.bridgeJSTypeID,
+        UInt.bridgeJSTypeID,
+        Int64.bridgeJSTypeID,
+        UInt64.bridgeJSTypeID,
+        Float.bridgeJSTypeID,
+        Double.bridgeJSTypeID,
+        String.bridgeJSTypeID,
+        JSValue.bridgeJSTypeID,
```
kateinoigakukun: Can we put those core type table entries in JavaScriptKit instead of defining in every module?

Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
```diff
@@ -138,6 +148,271 @@ extension JSGlueVariableScope {
...
+            // codec literal built from the same element stack fragments.
+            let codecVar = context.scope.variable("elemCodec")
+            try writeCodecLiteral(
+                type: type,
+                into: context.printer,
+                context: context,
+                prefix: "const \(codecVar) = ",
+                suffix: ";"
+            )
+            return codecVar
```
kateinoigakukun: Shouldn't we define a codec helper instead of inlining them? I think we can define StackABI codec helpers for any type that can appear in those combinators' element positions. And it should be shared with the one in type table entries.

Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSLinkTests/GenericImports.js
```diff
@@ -0,0 +1,936 @@
...
+            case 8:
+                jsValue = swift.memory.getObject(payload1);
+                break;
+            default:
+                throw new TypeError("Unsupported JSValue kind " + kind);
+        }
+        return jsValue;
+    }
+
+    const __bjs_createGenericPointHelpers = () => ({
```
kateinoigakukun: We need to include the module name to allow name collisions across modules.

krodak: Oh sorry, I reverted that change, will bring it back in a few 🙏🏻

Plugins/BridgeJS/Tests/BridgeJSToolTests/__Snapshots__/BridgeJSCodegenTests/Alias.swift
krodak: Fixed and moved 🙏🏻

Plugins/BridgeJS/Sources/BridgeJSLink/JSGlueGen.swift
krodak: Unified to shared codec helpers 🙏🏻

kateinoigakukun (approved):
Thanks! Feel free to merge once CI passed

PR #800, opened 2026-08-11
BridgeJS: Add runtime test for instance method String returns on a struct

wfltaylor:
Follow up to https://github.com/swiftwasm/JavaScriptKit/pull/789 to address @kateinoigakukun’s comment


kateinoigakukun (approved):
Thanks!

PR #803, opened 2026-08-14
BridgeJS: Allow extensions to contain types

wfltaylor:
Currently, you can write this:
```swift
@JS enum Change {
    case update(Update)
    case delete

    @JS struct Update {
        let contents: String
    }
}
```
but not this (which actually emitted broken generated code):
```swift
@JS enum Change {
    case update(Update)
    case delete
}

extension Change {
    @JS struct Update {
        let contents: String
    }
}
```

This PR removes this limitation, so you can define `@JS` types in extensions.

### DIFF
--- a/Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
+++ b/Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
@@ -139,6 +139,39 @@ public final class SwiftToSkeleton {
         sourceFiles.append((sourceFile, inputFilePath))
     }
 
+    private func resolveDeferredExtensions(_ exportCollectors: [ExportSwiftAPICollector]) {
+        var pendingExtensions = exportCollectors.flatMap { collector in
+            collector.deferredExtensions.map { (owner: collector, declaration: $0) }
+        }
+        var previousCount: Int
+        // An extended type might be defined in another extension, so keep resolving until no more progress is made.
+        repeat {
+            previousCount = pendingExtensions.count
+            var nextPendingExtensions: [(owner: ExportSwiftAPICollector, declaration: ExtensionDeclSyntax)] = []
+            for pending in pendingExtensions {
+                if !resolveExtension(pending.declaration, in: exportCollectors) {
+                    nextPendingExtensions.append(pending)
+                }
+            }
+            pendingExtensions = nextPendingExtensions
+        } while pendingExtensions.count < previousCount
+        for pending in pendingExtensions {
+            pending.owner.diagnoseUnresolvedExtension(pending.declaration)
+        }
+    }
+
+    private func resolveExtension(
+        _ declaration: ExtensionDeclSyntax,
+        in exportCollectors: [ExportSwiftAPICollector]
+    ) -> Bool {
+        for collector in exportCollectors {
+            if collector.resolveExtension(declaration) {
+                return true
+            }
+        }
+        return false
+    }
+
     public func finalize() throws -> BridgeJSSkeleton {
         var perSourceErrors: [(inputFilePath: String, errors: [DiagnosticError])] = []
         var importedFiles: [ImportedFileSkeleton] = []
@@ -244,9 +277,7 @@ public final class SwiftToSkeleton {
         }
 
         // Resolve extensions against all collectors. This needs to happen at this point so we can resolve both same file and cross file extensions.
-        for source in exportCollectors {
-            source.resolveDeferredExtensions(against: exportCollectors)
-        }
+        resolveDeferredExtensions(exportCollectors)
 
         // We have to collect diagnostics after all deferred extensions are resolved, since they could generate some.
         for ((_, inputFilePath), exportCollector) in zip(sourceFiles, exportCollectors) {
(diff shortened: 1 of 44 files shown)


wfltaylor:
I’ve pushed a commit to fix an existing .d.ts code-generation issue that the new tests triggered.

Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
```diff
@@ -1704,6 +1704,24 @@ public struct BridgeJSLink {
                 }
             }
             return type.tsType
+        case .swiftStruct(let name):
+            for skeleton in exportedSkeletons {
+                for structDef in skeleton.structs {
+                    if structDef.name == name || structDef.swiftCallName == name {
+                        return structDef.tsFullPath
```
krodak, line 1711: Supporting bare names alongside qualified names makes sense for older skeletons. Could we prefer `swiftCallName` before falling back to `name`?

Right now declaration order can select the wrong struct. With `Library.Shelf` before a top-level `Shelf`, `takeShelf(_:)` and even the top-level initializer are emitted as using `Library.Shelf`.

A snapshot with nested and top-level structs sharing a name would cover this.

```swift
case .swiftStruct(let name):
    let structs = exportedSkeletons.flatMap(\.structs)
    let match =
        structs.first { $0.swiftCallName == name }
        ?? structs.first { $0.name == name }

    return match?.tsFullPath ?? type.tsType
```

Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
```diff
@@ -1704,6 +1704,24 @@ public struct BridgeJSLink {
...
+                    }
+                }
+            }
+            return type.tsType
+        case .swiftHeapObject(let name):
+            for skeleton in exportedSkeletons {
```
krodak, line 1717: The new resolver fixes direct references to nested classes and structs, but closure signatures still use `BridgeType.tsType` without the same resolution.

I tried a callback taking and returning `Workshop.Bench`. The declaration uses `Workshop.Bench` inside the callback even though only the top-level `Bench` interface exists.

Could closure parameters and returns go through `resolveTypeScriptType` as well? Extending `NamespacedClassSignature` with a callback would cover it.

```swift
case .closure(let signature, _):
    let parameters = signature.parameters.enumerated().map { index, parameter in
        let type = resolveTypeScriptType(parameter, exportedSkeletons: exportedSkeletons)
        return "arg\(index): \(type)"
    }.joined(separator: ", ")

    let returnType = resolveTypeScriptType(
        signature.returnType,
        exportedSkeletons: exportedSkeletons
    )
    let resolvedReturnType =
        signature.isAsync ? "Promise<\(returnType)>" : returnType

    return "(\(parameters)) => \(resolvedReturnType)"
```

Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
```diff
@@ -1921,25 +1944,23 @@ private final class ExportSwiftAPICollector: SyntaxAnyVisitor {
...
     /// Walks extension members under the matching type’s state, returning whether the type was found.
-    ///
-    /// Note: The lookup scans dictionaries keyed by `makeKey(name:namespace:)`, matching only by
-    /// plain name. If two types share a name but differ by namespace, `.first(where:)` picks
-    /// whichever comes first. This is acceptable today since namespace collisions are unlikely,
-    /// but may need refinement if namespace-qualified extension resolution is added.
     func resolveExtension(_ ext: ExtensionDeclSyntax) -> Bool {
-        let name = ext.extendedType.trimmedDescription
+        guard let extendedDecl = parent.typeDeclResolver.resolve(ext.extendedType) else {
+            return false
```
krodak, line 2000: Deferred extension resolution works well for cross-file declarations, but a single pass leaves it dependent on source order.

I put `extension Library.Shelf` before the `extension Library` that declares `Shelf`. The generated struct was present, but the earlier extension's method was missing. Reversing the declarations restored it.

Could we retry unresolved extensions until a pass makes no progress? A reversed-order cross-file test would cover the same case across input files.

```swift
var pending = exportCollectors.flatMap { collector in
    collector.deferredExtensions.map { (owner: collector, extension: $0) }
}

repeat {
    let previousCount = pending.count
    pending.removeAll { item in
        exportCollectors.contains { $0.resolveExtension(item.extension) }
    }
    if pending.count == previousCount {
        break
    }
} while !pending.isEmpty

for item in pending {
    item.owner.diagnoseUnresolvedExtension(item.extension)
}
```

Plugins/BridgeJS/Sources/BridgeJSCore/TypeDeclResolver.swift
```diff
@@ -132,15 +143,15 @@ class TypeDeclResolver {
...
     ///
     /// - Parameter type: The SwiftSyntax node representing a type appearance in source code.
     /// - Returns: The nominal declaration (enum/class/actor/struct) if found, otherwise nil.
     func resolve(_ type: TypeSyntax) -> TypeDecl? {
         if let id = type.as(IdentifierTypeSyntax.self) {
             return lookupType(for: id)
         }
-        if let components = qualifiedComponents(from: type) {
+        if let components = type.qualifiedComponents {
             return lookupType(fullyQualified: components)
```
krodak, line 155: FYI, extension scoping now handles unqualified names, but relative qualified names such as `Shelf.Divider` are still treated as root-qualified. An `@JS` method using that spelling is omitted from the skeleton instead of resolving it as `Library.Shelf.Divider`.

This behavior already exists before this PR, so I opened #805 to follow it separately rather than treating it as part of this change.

krodak (commented):
Three inline findings. The pre-existing relative qualified-name issue is tracked separately in #805.

Plugins/BridgeJS/Sources/BridgeJSCore/SwiftToSkeleton.swift
wfltaylor: Addressed in the latest commit

Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
wfltaylor: Addressed in the latest commit, but I think we should probably followup to tidy this up. I don’t think the distinction between `tsType` and `resolveTypeScriptType` is clear enough, in fact maybe we could eliminate `tsType` since it’s usually not really what is wanted?

Plugins/BridgeJS/Sources/BridgeJSLink/BridgeJSLink.swift
wfltaylor: This was copied from above but I’m not sure it makes sense to match on the `.name` at all? I’ve eliminated this and now only match on the `swiftCallName`.

krodak (approved):

Plugins/BridgeJS/Sources/BridgeJSCore/TypeDeclResolver.swift
krodak: I'll open up PR for #805 after this one is merged

